PR #311, opened 2026-07-06
fix(app): register open_external_url on desktop and route by URL pattern

sonegillis1:
## Problem

On the **desktop** app, `openExternalUrl` failed with `Command open_external_url not found` and fell back to `window.location.href`, loading the URL **inside the main WebView**.

**Root cause:** the Tauri backend has two builders — `lib.rs` (mobile entry) registered `open_external_url`, but `main.rs` (`fn main()`, the **desktop** entry) did not, and didn't define it at all.

## Behavior

Define and register the command on desktop, and **route by URL pattern**:

- URLs matching `IN_APP_URL_PATTERNS` (currently **`stripe.com`**) open in an **in-app popup window** (reusing `open_oauth_in_popup`), matching the SSO login experience — these flows rely on the app session. Each entry carries its own popup **title** (Stripe → "Checkout", OAuth → "Sign In"); add more patterns/titles in that one list.
- **Everything else** opens in the **system browser** via the opener plugin.

**Return handling:** the popup's navigation handler now also detects a return to the app's own domain (`get_app_url()`, e.g. `os.ibl.ai/...`). When an in-app flow like Stripe checkout redirects back to the app, the popup **closes** and the **main window** navigates to that URL. The match is boundary-checked so lookalike hosts (`os.ibl.ai.evil.com`) don't trigger it.

Window creation for the popup is dispatched to the main thread (`run_on_main_thread`), since it's an explicit `invoke` and never hits the main window's `on_navigation` interceptor.

## Verification

- `cargo check` passes (the only warning — `unused import: tokio::sync::oneshot` — is pre-existing in `lib.rs`).

## Notes

- Requires a **desktop rebuild** to take effect (the fix is in the Rust binary).

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -30,6 +30,9 @@ use tauri::{command, AppHandle, Emitter, Listener, Manager, Window};
 use tokio::sync::RwLock;
 
 use tauri::WebviewWindowBuilder;
+// OpenerExt is used to open external URLs in the system browser (non-iOS).
+#[cfg(not(target_os = "ios"))]
+use tauri_plugin_opener::OpenerExt;
 
 // OAuth provider URL patterns that should be opened in a separate auth window
 // Note: login.iblai.app loads in the main window - only OAuth providers need a separate window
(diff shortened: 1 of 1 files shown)


mamigot:
**Claude finished @sonegillis1's task** —— [View job](https://github.com/iblai/os/actions/runs/28827593878)

---
### E2E Coverage Review

- [x] Dismiss stale REQUEST_CHANGES reviews
- [x] Identify changed files
- [x] Run mechanical coverage check

### Result

**CASE E — No new routes, coverage intact.**

The only changed file is `src-tauri/src/main.rs` (a Rust backend fix). No new `app/**/page.tsx` routes were added. The mechanical coverage check passed with **499 checkpoints** and **25/25 files mapped (100%)**.

A GitHub review was posted: ✅ **Approved**

mamigot (approved):
## ✅ E2E Coverage — No Issues

PR #312, opened 2026-07-06
feat(app): route open_external_url to an in-app popup by URL pattern

sonegillis1:
## Summary

Follow-up to #311, which registered `open_external_url` on desktop but always opened the **system browser**. Some flows (e.g. **Stripe checkout**) rely on the app session and should stay **inside the app**, like SSO login.

## Behavior

Route by URL:

- URLs matching `IN_APP_URL_PATTERNS` (currently **`stripe.com`**) open in an **in-app popup window** (reusing `open_oauth_in_popup`). Each entry carries its own popup **title** (Stripe → "Checkout", OAuth → "Sign In"); add more patterns/titles in that one list.
- **Everything else** keeps opening in the **system browser**.

**Return handling:** the popup's navigation handler now detects a return to the app's own domain (`get_app_url()`, e.g. `os.ibl.ai/...`). When an in-app flow like Stripe checkout redirects back to the app, the popup **closes** and the **main window** is navigated to that URL. The match is boundary-checked so lookalike hosts (`os.ibl.ai.evil.com`) don't trigger it.

The handoff uses `WebviewWindow::navigate()` (a hard load) rather than assigning `window.location.href`, so the Next.js SPA router can't swallow the navigation and leave the main window on the page that opened the popup. The OAuth callback path keeps using `eval` (unchanged).

Popup window creation is dispatched to the main thread (`run_on_main_thread`), since it's an explicit `invoke` and never hits the main window's `on_navigation` interceptor.

## Verification

- `cargo check` passes (the only warning — `unused import: tokio::sync::oneshot` — is pre-existing in `lib.rs`).

## Notes

- Requires a **desktop rebuild** to take effect (the fix is in the Rust binary).

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -30,7 +30,8 @@ use tauri::{command, AppHandle, Emitter, Listener, Manager, Window};
 use tokio::sync::RwLock;
 
 use tauri::WebviewWindowBuilder;
-// OpenerExt is used to open external URLs in the system browser (non-iOS).
+use tokio::sync::oneshot;
+// OpenerExt opens URLs in the system browser (non-iOS).
 #[cfg(not(target_os = "ios"))]
 use tauri_plugin_opener::OpenerExt;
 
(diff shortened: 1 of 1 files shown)


mamigot:
**Claude finished @sonegillis1's task** —— [View job](https://github.com/iblai/os/actions/runs/28831010609)

---
### E2E Coverage Check

- [x] Identify changed files — only `src-tauri/src/main.rs` (no new `page.tsx` routes)
- [x] Run mechanical coverage check — passed (499 checkpoints, 25/25 files mapped)
- [x] Read current coverage state — all existing coverage intact
- [x] Applied decision logic: **CASE E** — no new app routes, coverage check passed → approved

## ✅ E2E Coverage — No Issues

This PR only modifies `src-tauri/src/main.rs` (Rust/Tauri backend). No new frontend `page.tsx` routes were added. The mechanical coverage check passed with 499 checkpoints and 100% route coverage.

mamigot (approved):
## ✅ E2E Coverage — No Issues

PR #317, opened 2026-07-07
fix(app): foreground main window before navigating on popup return

sonegillis1:
## Summary

Follow-up to #314. The `navigate()` handoff works, but after the in-app popup closed there was a **noticeable delay** before the main window loaded the return URL.

## Cause

The handler navigated the main window **while it was still behind the popup**. A backgrounded WebView is heavily throttled on macOS, so the navigation/load was deferred until the main window came to the foreground (after the popup closed).

## Fix

Reorder the handoff so the main window is already foreground when it navigates:

1. Compute the target URL.
2. **Close the popup.**
3. **Focus the main window.**
4. **Then** navigate it (`navigate()` for app-domain returns, `eval` fallback / OAuth path).

No behavior change beyond timing — the same URL is loaded, just without the background-throttle delay.

## Verification

- `cargo check` passes.
- Requires a **desktop rebuild** to take effect.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -123,17 +123,27 @@ fn open_oauth_in_popup(url: &str, app_handle: &AppHandle, title: &str) -> Result
         if is_callback || is_custom_scheme || is_app_return {
             println!("[OAuth Popup] Callback/return detected: {}", url_str);
 
-            // Navigate the main window to the callback URL
+            let target_url = if is_custom_scheme {
+                // Convert custom scheme to app URL
+                let path = url_str
+                    .replace("iblai-mentor://", "/")
+                    .replace("ai.ibl.mentorai://", "/");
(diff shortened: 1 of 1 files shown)


mamigot:
**Claude finished @sonegillis1's task** —— [View job](https://github.com/iblai/os/actions/runs/28886240214)

---
### E2E Coverage Review

- [x] Dismiss stale REQUEST_CHANGES reviews
- [x] Identify changed files
- [x] Run mechanical coverage check
- [x] Read current coverage state

### Result: ✅ E2E Coverage — No Issues

Only `src-tauri/src/main.rs` was changed — no new frontend routes (`page.tsx`) were introduced. The mechanical coverage check passed:
- **503 checkpoints** — no regression
- **25/25 files mapped (100%)** — all routes covered

PR approved.

mamigot (approved):
## ✅ E2E Coverage — No Issues

PR #333, opened 2026-07-13
fix(auth): require a valid edx_jwt_token for a non-expired session

sonegillis1:
## Problem

`hasNonExpiredAuthToken()` only validated the **axd** token. If the **edx JWT** (`edx_jwt_token`) was missing or expired, the session was still treated as live and re-authentication was skipped — even though the edx token is set alongside axd at SSO login and is required for a working session.

## Change

`hasNonExpiredAuthToken()` now also requires `edx_jwt_token` to be **present and unexpired** before returning `true`. The check runs first, so every success path is gated on it; when the edx token is missing or expired, the function returns `false` and the SDK's `redirectToAuthSpa` re-authenticates.

- New `isJwtExpired(token)` helper decodes the JWT's `exp` claim (base64url-safe). Malformed/undecodable → treated as expired; no `exp` claim → treated as non-expiring (mirrors the existing axd "no expiry → valid" behavior).
- Semantics are **AND**: both axd and edx must be valid.

## Tests

`lib/__tests__/utils.test.ts`: seeds a valid edx JWT in `beforeEach` (so the axd assertions still exercise axd logic), adds edx cases (missing / expired / both-valid), and a full `isJwtExpired` describe (future/past/no-exp/no-payload/undecodable). **339 passed** locally; typecheck clean.

## Note

Pushed with `--no-verify` (author-authorized): the full pre-push suite passed, but `git push` hit a repeated SIGPIPE-after-pass transport flake in the hook. The pre-existing `###################` debug `console.log`s in this function were left as-is (matched style for the new branches); happy to strip them all in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,13 @@
 
 ### CI
 
-* **pr-gate:** run vitest unit tests in the Required PR workflow ([dbea82a](https://github.com/iblai/os/commit/dbea82a2bf35529c6195fb8ab1b631d2b2b09ce4))
+- **pr-gate:** run vitest unit tests in the Required PR workflow ([dbea82a](https://github.com/iblai/os/commit/dbea82a2bf35529c6195fb8ab1b631d2b2b09ce4))
 
 ## [0.95.6](https://github.com/iblai/os/compare/v0.95.5...v0.95.6) (2026-07-13)
 
 ### Documentation
 
-* **readme:** link the macOS row to the latest build download ([238bc6d](https://github.com/iblai/os/commit/238bc6d16346989f0fc1f4a52542918f8c8bbab3)), closes [#download](https://github.com/iblai/os/issues/download)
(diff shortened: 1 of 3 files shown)


mamigot:
**Claude finished @sonegillis1's task** —— [View job](https://github.com/iblai/os/actions/runs/29294045830)

---
### E2E Coverage Review

- [x] Identify changed files (`CHANGELOG.md`, `lib/__tests__/utils.test.ts`, `lib/utils.ts` — no new `page.tsx` routes)
- [x] Run mechanical coverage check (✅ exit code 0, 545 checkpoints, 25/25 files mapped)
- [x] Read current coverage state (all intact, no regression)
- [x] Apply decision logic — **CASE E**: No new routes, mechanical check passed → Approved

**Result:** PR approved. No new app routes were introduced; all existing coverage is intact.

mamigot (approved):
## ✅ E2E Coverage — No Issues

**Coverage check:** 545 checkpoints, 25/25 files mapped (100%)

PR #353, opened 2026-07-20
fix(tenant-switch): complete auth handoff + pull in SDK loader-hang fix (iblai-js 1.26.3)

sonegillis1:
## Summary

Fixes the intermittent bug where a tenant switch never completes and the app stays on the loading screen. Two parts: the SDK fix (via version bump) and the app-side SSO handoff.

## Changes

- **`@iblai/iblai-js` → 1.26.3**, which carries the `web-utils` `AuthProvider` fix ([ibl-frontend#1932](https://github.com/iblai/iblai-web-frontend/pull/1932)): when a tenant switch lands back on the app before the 8s switch lock expires, `performAuthCheck` used to bail without ever re-running, latching `isAuthenticating` true → infinite loader. It now self-schedules a re-check past the lock TTL.
- **Parse the SSO handoff payload** in the iframe `defaultHandler`: the payload can arrive as a JSON **string** (cross-origin `postMessage`) rather than an object, so `axd_token` was never detected and the new session wasn't saved. Now normalized with a `JSON.parse` guard.
- **`redirectToMentor`** preserves the current query string on **every** path, not only `/`, so params survive the post-switch redirect into the mentor.

## Validation

Full pre-push hook passed green — build, unit suite, per-file coverage (`providers/index.tsx` is in the coverage skip-list), and e2e-journey coverage. `pnpm typecheck` clean against 1.26.3.

The final push used `--no-verify` only to work around an environmental git SIGPIPE that kills the upload after the hook completes cleanly (recurring on this machine); the hook itself was validated green.


mamigot:
**Claude finished @sonegillis1's task** —— [View job](https://github.com/iblai/os/actions/runs/29778440687)

---
### E2E Coverage Check

- [x] Identify changed files (`package.json`, `pnpm-lock.yaml`, `providers/index.tsx`)
- [x] Run mechanical coverage check — **passed** (555 checkpoints, 25/25 files covered)
- [x] Read current coverage state

**Result:** No new app routes introduced. All existing coverage intact (555 checkpoints, 0 regression).

mamigot (approved):
## ✅ E2E Coverage — No Issues

PR #379, opened 2026-07-28
fix(tauri): end intermittent app freeze from webview-manager mutex deadlock

sonegillis1:
## The freeze

The desktop app intermittently froze ("Not Responding"). A `sample` of a frozen instance captured a **two-thread mutex deadlock** on Tauri's webview manager:

- **Main (UI) thread** — `handle_ipc_message → AppManager::get_webview → Mutex::lock`, servicing a webview IPC message.
- **tokio worker** — `check_ollama_status → AppHandle::emit(OllamaStatus) → Emitter::emit → Mutex::lock`, fanning an event out to every webview.

Tauri's app-wide `Emitter::emit` locks the webview-manager mutex. Every async `#[command]` here runs on a tokio worker, so calling `app.emit(...)` directly from one races the main thread locking the **same** mutex to look up a webview. When they collide, the app deadlocks — and it was intermittent because it needs simultaneous lock contention. `check_ollama_status` (polled by the model-download UI) plus the streaming token emits made that collision likely.

## The fix

Add an `emit_on_main` helper that marshals every emit onto the main thread via `run_on_main_thread`, so emits **serialize** with IPC handling and can never contend the manager mutex across threads. `run_on_main_thread` only posts to the event loop (non-blocking, FIFO), so streamed events (`ollama:token`, …) stay ordered.

Converted every `app.emit(...)` in the async commands of:
- `src-tauri/src/main.rs` — the **desktop** binary entry (the frozen Mac app).
- `src-tauri/src/lib.rs` — the **mobile** entry point, hand-kept in sync, which had the identical deadlock.

## Verification

- `cargo check` passes (both `bin` and `lib` targets).
- Pre-push validation (typecheck, lint, build, 10560 unit tests, coverage) passed green — Rust-only change, no JS behaviour touched. Landed via authorized `--no-verify` due to the known mentorai SIGPIPE-after-green-hook transport flake.

## Follow-up (optional, not in this PR)

`check_ollama_status` is polled frequently by `useModelDownload`/`useTauri`. Throttling that render storm would cut redundant main-thread emit traffic. Not required for correctness now that emits can't deadlock, but a worthwhile cleanup.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -781,6 +781,31 @@ async fn stop_ollama(app: AppHandle) -> Result<(), String> {
     Ok(())
 }
 
+/// Emit an app-wide event to the frontend WITHOUT racing the main thread's
+/// webview-manager mutex.
+///
+/// Tauri's app-wide `Emitter::emit` locks the webview-manager mutex to fan a
+/// payload out to every webview. Every async `#[command]` here runs on a tokio
+/// worker, so calling `app.emit(...)` directly from one can deadlock against the
+/// main thread locking the SAME mutex to service a webview IPC message
+/// (`AppManager::get_webview`) — the intermittent "Not Responding" freeze
+/// (see `check_ollama_status`, which the UI polls). Marshalling every emit onto
+/// the main thread serializes it with IPC handling, so the two can never contend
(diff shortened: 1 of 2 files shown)


mamigot:
**Claude encountered an error** —— [View job](https://github.com/iblai/os/actions/runs/30407510243)

---

PR #383, opened 2026-07-29
feat(ci): use stg2 as overflow when stg1 is busy

bnsoni:
PR e2e always targeted **stg1**, so two concurrent PRs serialised while stg2 sat idle. This picks the environment just before dispatch.

## Policy
**stg1 primary · stg2 overflow — only when stg1 is busy AND stg2 is free.**

| stg1 | stg2 | → | why |
|---|---|---|---|
| free | — | **stg1** | primary |
| busy | free | **stg2** | fail over |
| busy | busy | **stg1** | queue on primary |
| busy | not seen | **stg1** | stg2 not a live runner |

The both-busy case matters: queueing on stg2 would mean waiting behind **release validation** and then downgrading that box to the prod tag, forcing validation to redeploy. So it queues on stg1 instead.

## Why this needs no runner labels or a lock
`pr-e2e` serialises on `env-mutate-<env>`. If two PRs race and both pick stg1, the second simply **queues** — todays behaviour. A wrong guess costs throughput, never correctness. That is what makes a best-effort check sufficient here.

## How busy is detected
`runner_name` on the jobs API — the precise signal for "is that box executing work", and it catches *any* workflow occupying the runner, not just ones whose name mentions the env. Failover additionally requires having *seen* `stg2-spa` execute work recently, so an offline box cannot swallow the run into a queue that never drains.

## Verification
- executed against **live** deploy-ops state mid-smoke-test → correctly returned `stg2 (stg1 busy, stg2 free — failing over)`, since stg1-spa was genuinely occupied
- all five decision branches exercised under bash
- caught a real fragility on the way: the first version relied on unquoted word-splitting, which is bash-only and silently returns the wrong answer elsewhere. Now splits explicitly with `tr`.

### DIFF
--- a/.github/workflows/pr-e2e-tests.yml
+++ b/.github/workflows/pr-e2e-tests.yml
@@ -128,6 +128,47 @@ jobs:
             echo "full suite (changed specs=$N, non-spec changes=${OTHER:-none})"
           fi
 
+      - name: Pick a target environment
+        id: env
+        env:
+          GH_TOKEN: ${{ secrets.DEPLOY_OPS_DISPATCH_TOKEN }}
+        run: |
+          set -uo pipefail
+          # stg1 is PRIMARY; stg2 is OVERFLOW, used only while stg1 is genuinely occupied. That keeps
+          # release validation — which lives on stg2 — undisturbed in the common case, while still
+          # letting two concurrent PRs use both boxes.
+          #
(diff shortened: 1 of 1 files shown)


mamigot:
**Claude encountered an error** —— [View job](https://github.com/iblai/os/actions/runs/30419530222)

---

PR #380, opened 2026-07-29
ci(e2e): run PR e2e on the central pipeline (replaces OCI container instances)

bnsoni:
## What
Replaces the OCI container-instance e2e path with the **central pipeline** in `iblai/iblai-deploy-ops` — same style, gate and report as the stg2 release-validation flow.

## Flow (unchanged trigger: the `run-tests` label)
```
run-tests label → build the PR SPA image → dispatch iblai-deploy-ops/pr-e2e.yml
                → central pipeline: swap the env SPA to this PR image, build the Playwright
                  image from THIS PR sha, run the suite, publish the report, ALWAYS restore
                → this job polls and mirrors the result
```
Because the `e2e` job carries the real conclusion, **`PR E2E Tests (Required)` still works as a required check** and blocks merge until e2e passes. The workflow `name:`, trigger, path filters and concurrency are unchanged.

## Improvements over the OCI path
- **The env is restored afterwards.** The old flow `sed`-swapped the SPA image and never put it back — which is why stg1 was found still serving `iblai-mentor-spa-pro:pr-371-…` days later. Restore now runs `if: always()`.
- **Changed-spec detection** — if a PR only touches spec files (≤5), just those run (minutes instead of ~1h); any app-source change runs the full suite.
- **Real reports** — the standard S3/CloudFront report, regression-only gate, traces and 30-day history, instead of raw logs.
- **One less image build** — the central pipeline builds the Playwright image from the PR sha, so `build-playwright-image` is gone.
- Runs on the **stg1** environment via the maintained runner pool, so PR testing no longer competes with release validation on stg2.

## Safety (unchanged)
Both guards kept: the `run-tests` label (labels need write access, so a maintainer vouches) **and** `head.repo.full_name == github.repository` (no fork PRs). Dispatch+poll is required because a public repo cannot call a private reusable workflow.

## Requires
Org secret `DEPLOY_OPS_DISPATCH_TOKEN` (fine-grained PAT, Actions: read+write on `iblai-deploy-ops`) — already created and scoped to this repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #372, opened 2026-07-27
feat(security): enforce CSP by default (report-only in dev)

sonegillis1:
## What

Flips the nonce-based Content-Security-Policy from **report-only** to **enforced by default in production**. Development stays report-only so `next dev`'s HMR websocket, eval-based React Refresh, and the error overlay aren't blocked. `CSP_MODE` still overrides either way (`report-only` | `enforce`).

```
isEnforce() = CSP_MODE==='enforce'  ? true
            : CSP_MODE==='report-only' ? false
            : NODE_ENV !== 'development'   // enforce in prod, report-only in dev
```

## Why

Report-only provided **no actual protection** — it evaluated the policy and logged violations but blocked nothing. Enforcement is what makes the CSP mitigate XSS/script-injection: only nonce-carrying (and `strict-dynamic`-propagated) scripts run, and resources are constrained to the allowlist. The policy itself is unchanged and has been live in report-only.

## Tested (production build via `next start`, headless browser)

- **Header flips:** responses carry an enforcing `Content-Security-Policy` (not `-Report-Only`), including `upgrade-insecure-requests`.
- **App boots under enforcement:** 112 nonced Next.js chunks execute (`strict-dynamic` + nonce verified), body renders.
- **Zero `securitypolicyviolation` events** across the embed mentor page, `/`, and `/sso-login`.
- **Enforce ≡ report-only:** identical behavior on the same build — same chunks, same render; the only failed requests are backend network aborts (`api.iblai.app` / `sentry.ibl.network`) that occur in *both* modes and are **not** CSP (allowed by `connect-src`; no violation events). So flipping the default blocks nothing legitimate.

Unit tests: 10 pass, **100% line coverage** on `middleware.ts` (enforce-default, dev report-only, both `CSP_MODE` overrides). typecheck ✓, lint ✓, build ✓, e2e coverage unchanged.

## Rollout notes

- **Ops:** the static upstream (Nginx) `Content-Security-Policy` header MUST be removed on deploy — browsers intersect multiple CSP headers, so a stale `default-src 'self'` would block the nonced scripts under enforcement. Keep static HSTS / X-Content-Type-Options / Referrer-Policy in Nginx.
- To validate a new third-party origin later without risk, set `CSP_MODE=report-only` on that environment, observe, then remove the override.
- Wiring `CSP_REPORT_URI` to a Sentry CSP endpoint (as done in the auth app) is still recommended to catch any environment-specific violations post-enforcement.

[...]

PR #384, opened 2026-07-29
chore: merge main into feat/csp-enforce-default (pick up the new PR e2e pipeline)

bnsoni:
Merges current `main` into `feat/csp-enforce-default` so that **#372 runs the new central e2e pipeline**. No changes to the branch's own work — this only brings it up to date.

### Why it is needed
A `pull_request` workflow runs from the **merge commit**, and #372's merge ref still carries the old workflow:

| ref | workflow |
|---|---|
| `refs/pull/372/merge` | **154 lines** — old OCI runner, no `merge_sha` |
| `refs/heads/main` | **311 lines** — new central pipeline |

The branch is **31 commits behind** main, from before `#380` replaced the caller. Labelling #372 today would fire the **old** pipeline.

### Why that matters beyond just "old"
The old OCI flow swapped stg1's SPA with a bare `sed` and **never restored it**. That is how stg1 was found still serving `iblai-mentor-spa-pro:pr-371-…` days later, during an unrelated 1.110.0 deploy. Running it again would leave stg1 on this PR's build.

### What #372 gets instead
- tested against the release **running in production**, with the env synced to it first if it has drifted
- the SPA image and the Playwright suite both built from the PR's **merge commit** — no version skew
- the SPA **restored** afterwards, always
- live phase progress in this repo's own log, and a PR comment with a public link to the report and traces

### Conflicts
The branch does not touch `.github/workflows/pr-e2e-tests.yml`, so the file that matters merges cleanly.

### After merging
`run-tests` is already applied to #372, so the merge fires `synchronize` and the run starts automatically — no need to re-label.

Note this is also the situation for **every** currently-open os/lms PR: they will keep running the old pipeline until their branches move past the merge of `#380`.


bnsoni:
Superseded by a minimal version that changes only the workflow file — merging all of `main` (30 files) into the branch was more than needed. The caller's only local dependency, `reusable-pr-docker-build.yml`, is byte-identical on both refs, so one file is sufficient.

PR #377, opened 2026-07-28
Fix/2260

conradmugabe:
## Checklist
- [x] Tests were added/updated according to the feature/bugfix/change made
- [x] Version was rolled according to [semver requirements](https://semver.org/#summary)
- [x] API endpoints openapi schema was updated if applicable

## Changes
- Make long proactive prompt scrollable

## Screenshots
Screenshot from 2026-07-28 21-20-57
Screenshot from 2026-07-28 21-20-47
Screenshot from 2026-07-28 21-21-25
Screenshot from 2026-07-28 21-21-08

PR #386, opened 2026-07-29
ci: adopt the central PR e2e pipeline on this branch

bnsoni:
One file. Lets **#377** run the new central e2e pipeline instead of the old OCI runner.

### Why it is needed
A `pull_request` workflow runs from the **merge commit**, and #377's merge ref still carries the pre-#380 caller:

| ref | workflow |
|---|---|
| `refs/pull/377/merge` | **154 lines** — old OCI runner |
| `refs/heads/main` | **311 lines** — new central pipeline |

### Why it matters
The old OCI flow swapped stg1's SPA with a bare `sed` and **never restored it** — that is how stg1 was found still serving `iblai-mentor-spa-pro:pr-371-…` days later, during an unrelated 1.110.0 deploy. Running #377 on it would leave stg1 on this branch's build.

### Scope — deliberately minimal
Only `.github/workflows/pr-e2e-tests.yml`, copied from `main`. **No** other code from main is pulled in; the branch's own work is untouched.

Sufficient because the caller's single local dependency, `reusable-pr-docker-build.yml`, is **byte-identical** on `main` and this branch (`e7e1f709332b`) and already accepts every input it passes.

### What #377 gets
- tested against the release **running in production**, with the env synced first if it has drifted
- SPA image and Playwright suite both built from the PR's **merge commit** — no version skew
- the SPA **restored** afterwards, always
- live phase progress in this repo's log, and a PR comment linking the report and traces

This PR touches `components/**` and an `e2e/` page object, so the changed-spec detector will select the **full** mentor suite (~50 min) rather than a selective run — expected for a change of this shape.

### After merging
`run-tests` is already on #377, so the resulting `synchronize` fires the run automatically — no re-labelling.

Same treatment as #385 (for #372).

### DIFF
--- a/.github/workflows/pr-e2e-tests.yml
+++ b/.github/workflows/pr-e2e-tests.yml
@@ -24,10 +24,11 @@ on:
       - 'vitest.config.*'
       - 'playwright.config.*'
 
-run-name: 'PR E2E Tests — stg1 (firefox,chrome,safari,edge)'
+run-name: 'PR E2E — mentor #${{ github.event.pull_request.number }} → stg1 (chrome)'
 
 permissions:
   contents: read
+  pull-requests: write   # the e2e job posts the report link as a PR comment
 
 concurrency:
   group: pr-e2e-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


mamigot:
**Claude encountered an error** —— [View job](https://github.com/iblai/os/actions/runs/30422181057)

---

PR #392, opened 2026-07-29
fix: repair main gates (iframe handler envelope reads + brace-expansion coverage crash)

sonegillis1:
Repairs two things that left `main`'s gates red and blocked pushes for everyone.

## 1. Broken pre-push coverage gate (`fix(deps)`)
The `brace-expansion: '>=5.0.8'` security override forced **every** brace-expansion in the workspace to v5 (ESM-style named exports). `minimatch@9` (glob → test-exclude → istanbul coverage) calls it as a CJS-callable default, so the local coverage check crashed deterministically:
```
TypeError: (0 , brace_expansion_1.default) is not a function
```
The tree has three minimatch majors with incompatible import styles (`@3`/`@9` need the callable line, `@10` needs the named `{ expand }` line), so no single pin works. Fix: default brace-expansion to the patched **2.x** callable line and **scope the patched 5.x to minimatch@10**. All lines stay past CVE-2025-5889.

## 2. iframe handlers read the wrong arg (`fix(iframe)` ×2)
The host posts `{ type, data }` via `postMessage`; the SDK dispatcher forwards the raw event (its `payload` destructure misses the `data` field). Three handlers read the wrong place:
- `MENTOR:ENABLE_GRADING` — dispatched the raw `MessageEvent` (always truthy, so `false` was ignored). Now reads `event.data.data`.
- `MENTOR:DOCUMENTFILTER` — `JSON.parse(event.data)` on the envelope object, threw + swallowed → filter never applied. Now `JSON.parse(event.data.data)`.
- `MENTOR:EDX_USAGE_ID` / `MENTOR:EDX_COURSE_ID` — dereferenced the undefined first arg, threw inside the SDK try/catch → edX metadata never set. Now read `event.data.data.*`.

Verified against the real senders in `agent-ai` (`src/index.ts`).

## Verification
- typecheck clean; full unit suite **10,577 pass**; `handlers.ts` coverage **97.06% ≥ 95%**.
- The pre-push gate now runs to completion (the coverage step no longer crashes).

> Landed via authorized `--no-verify` due to the known post-green-hook SIGPIPE-141 flake on this repo; the full gate was confirmed green beforehand.

### DIFF
--- a/.github/workflows/reusable-release-macos-dmg.yml
+++ b/.github/workflows/reusable-release-macos-dmg.yml
@@ -135,6 +135,73 @@ jobs:
         with:
           args: ${{ inputs.tauri_args }}
 
+      # Keep the product name "ibl.ai" but hide the ".app" suffix in the DMG's
+      # Finder label. Finder does NOT auto-hide ".app" for a dotted name like
+      # "ibl.ai.app" (it does for a clean name like "MentorAI.app") unless the
+      # bundle carries the per-item "extension hidden" FinderInfo bit. Tauri
+      # builds+signs+notarizes+staples the DMG in one shot with no hook between
+      # ".app built" and "DMG built", and codesign REJECTS FinderInfo set on the
+      # bundle *before* signing. So we set the bit *after* Tauri is done, then
+      # re-master the DMG (its layout/background survive `hdiutil convert`) and
+      # re-sign + re-notarize + re-staple it — the edit invalidates the DMG's own
+      # staple, but the inner .app's notarization is unaffected because the bit
+      # lives in the root FinderInfo, which the code signature does not seal.
+      - name: Hide .app extension in DMG Finder label
+        env:
+          # APPLE_SIGNING_IDENTITY / APPLE_TEAM_ID come from the job env.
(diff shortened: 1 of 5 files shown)


sonegillis1:
**Update — brace-expansion security correction (commit \`ba189b34\`).**

An earlier commit on this branch pinned `brace-expansion` to the `2.x` callable line to stop the istanbul-coverage crash. CI's `dependency-review` correctly flagged that: advisory **GHSA-mh99-v99m-4gvg** (DoS/OOM) is patched **only in `5.0.8`** — there is no patched 1.x/2.x/3.x/4.x.

Corrected approach (supersedes the 2.x pin):
- `brace-expansion: '>=5.0.8'` — the only patched version, applied everywhere.
- `minimatch@9: '^10'` — the lone dev-only consumer that couldn't read brace-expansion@5's named export (`@vitest/coverage-istanbul` → `glob@10` → `test-exclude`) is bumped to v10, which does.

Verified: lockfile carries only `brace-expansion@5.0.8`; `pnpm install --frozen-lockfile` clean; istanbul coverage runs to completion (395 files, 10,577 tests). `dependency-review` + `audit` now pass.

PR #415, opened 2026-08-05
feat(pr-e2e): call iblai/pr-e2e-action@v1 instead of an in-repo copy

bnsoni:
Replaces the 396-line `e2e` job with a call to **[iblai/pr-e2e-action@v1](https://github.com/iblai/pr-e2e-action)**. Ships together with the matching PRs in the other two caller repos.

## Why

A `pull_request` event runs workflows from the PR's **merge ref**, so this workflow is *frozen per PR*. A fix to it reaches **no open PR** until that PR rebases.

On 2026-08-04 that meant:
- a merged fix reached **none of ~27 open PRs** — each needed a rebase before it applied
- every bug needed **three PRs** (os + lms + web-frontend)
- different PRs ran different vintages of the pipeline

`uses: …@v1` is resolved **at run time from the action repo**, so a fix tagged `v1` applies to every PR immediately — including PRs opened months ago, with no rebase. That is the entire point of this change.

## What changed

Only the `e2e` job body. Everything repo-specific stays exactly where it is.

| job | |
|---|---|
| `gate` | unchanged — label + fork check remains the authorization model |
| `unit-tests` | unchanged *(os only)* |
| `build-app-image` | unchanged |
| **`e2e`** | **396 lines → a `uses:` step** |
| `summary` | unchanged — still the required-check shape |

**`name: E2E (central pipeline)` is preserved exactly** — it is the name branch protection matches on. Changing it would make every PR unmergeable. Same for `needs: build-app-image`.

## The action is a verbatim lift

Extracted from `os@main` including the comments (they carry the incident context for each guard), with every fix merged today already in it:

- the **any-self-hosted liveness probe** — a guessed `${TARGET}-spa` made a healthy run on the other box look offline and abandoned it 62 min before its report was written
- the **queue-count `wait_for_slot`** — title matching went blind when an env override changed the run-name, and PRs live-locked
- the **generic no-report summary** — a conclusion whitelist let `runner_unavailable` fall through to "ask a maintainer"
- the **carried-forward report link** — an evicted run used to overwrite a real verdict

Parameterised: the two hardcoded repo values plus every `secrets.` / `github.event.` / `needs.` reference became inputs. Audited: **0 leftovers** of each.

## Verification

- YAML parses in all three repos; `name` and `needs` asserted unchanged
- `summary` still resolves every `needs.*.result` it references; **0 orphaned step refs**
- All 6 embedded scripts in the action pass `bash -n`
- All 10 action inputs referenced, none unused

After merge, verify one PR per repo end to end — dispatch, phase streaming, **report link on the PR**, required check green — then confirm the propagation property by moving `v1` and watching an existing stale PR pick it up without rebasing.

## Rollout note

All three repos pin `@v1` (moving). A defect therefore affects all three at once — the trade is that recovery is one commit plus a tag move, instead of three PRs and ~27 rebases. `v1.0.0` is immutable, so `v1` can be rolled back instantly.


bnsoni:
## Review finding — fixed, please re-review

A defect was found reviewing this before merge, and it would have been a bad one to ship.

**The cancellation cleanup cannot live inside a composite action.** A composite action is
*terminated* when the job is cancelled, so the `if: cancelled()` step that cancels the dispatched
central run would never have fired.

That is the **routine path**, not an edge case: this job uses `cancel-in-progress: true` and a label
recycle fires two events, so the first caller is cancelled on essentially every re-tag. Shipping it
would have permanently re-introduced orphaned runs — each holding an environment for a full suite,
producing a result nobody reads, and sometimes **passing while the PR shows red** (os#394 did exactly
that). The step being moved was written to prevent precisely this.

### Fix

- the step is back in **this workflow as a job step**, where `if: cancelled()` is honoured
- the action (`v1.0.1`) now writes the run id to `$RUNNER_TEMP/pr-e2e-run-id` at dispatch time,
  because a composite action's declared outputs are only published when it **completes** — a
  cancelled action exposes no `run_id` at all

### Re-verified

YAML parses · `name: E2E (central pipeline)` and `needs` unchanged · `summary` still resolves ·
0 orphaned step refs · all 5 embedded action scripts pass `bash -n` · no `cancelled()` left inside
the action.

Action is now `v1.0.1`; `v1` points at it. `v1.0.0` remains immutable for rollback.

PR #464, opened 2026-08-28
Fix/2402

conradmugabe:
## Checklist

## Changes
- Reject empty blobs and recordings under 500ms before they reach the network; drop
  zero-length `dataavailable` chunks; pass a timeslice to `MediaRecorder` so the
  container header and first cluster flush early.
- Log blob size, mime type and duration on the failure path so the next occurrence is
  diagnosable.
- Guard `getUserMedia` and recorder startup — previously an unhandled rejection with no
  toast and a dead button on permission denial, no device, or an insecure context.
- Detect the mime type via `isTypeSupported` and read it back off the recorder instead
  of hardcoding `audio/webm`, which was mislabelling Safari's mp4 output.
- Ref guard so a second click during the permission prompt can't orphan a live stream.
- Stop tracks on unmount — navigating away mid-recording left the mic indicator on.
- Dictation appends to the composer instead of replacing it, so text typed before or
  during the recording survives.
 
 - Closes https://github.com/iblai/iblai-platform/issues/2402


michael-on-code (approved):
🟠 The recording timer never stops after recorder.onerror

[use-voice-chat.ts:191-197](vscode-webview://0fspfsluo7so688j9gp026tjve2rvhv38o9qo2d85flunqneu1n8/hooks/use-voice-chat.ts#L191-L197) — onerror calls stop(), but that stop was captured from the render in which startRecording ran, i.e. before start() flipped isRunning to true. useTimer.stop is a useCallback over [isRunning] and is a no-op when its captured isRunning is false:


const stop = useCallback(() => {
  if (isRunning) { … }   // false in the captured closure → nothing happens
}, [isRunning]);
Confirmed with a repro against the real useTimer (the unit tests mock the whole module, so they cannot see this). After onerror fires 2s into a recording:

the 10ms setInterval keeps running for the life of the component, re-rendering ChatInputForm 100×/s;
isRunning stays true, so the next start() is also a no-op — the next recording's placeholder at [chat-input-form.tsx:815](vscode-webview://0fspfsluo7so688j9gp026tjve2rvhv38o9qo2d85flunqneu1n8/components/chat-input-form.tsx#L815) reads Listening... 00:05 and keeps climbing from the stale origin. My repro measured 5100ms where 100ms was expected; the control (normal stop → restart) correctly reset to 100ms.
Fix: keep start/stop in refs (same pattern already used for onTranscript), or route onerror through a ref to the current stopRecording. Then unmock useTimer for one regression test — with the module mocked, mockStop is a stable vi.fn() and this class of bug stays invisible.

PR #472, opened 2026-09-02
feat(code-mode): three asked setup steps + guidance delivered as per-session AGENTS.md

null-crafter:
## Checklist

## What

  Two related changes to how the Code agent (opencode) is steered:

  **1. Web-app builds now walk three loose, asked steps** (in `IBLAI_INSTRUCTIONS`):
  - **Step 1 — the template question** (existing, now labeled): one short question, "our default template", never "vibe-starter".
  - **Step 2 — the local-preview question** (new): after building, ask whether the user wants a local preview at http://localhost:3000. Never start a dev server or open a browser at localhost without that yes; on yes, `pnpm dev` in the background + open the browser; hot-reload silently afterwards. This replaces the old "never show localhost or offer a local dev server" rule.
  - **Step 3 — the deploy question** (new): asked once per project ("put it on a live URL with our hosting"); yes = deploy now and auto-redeploy after later changes (hash-dedupe still skips no-ops), no = deploy only on request. Replaces the old auto-deploy-without-asking rule. typecheck/lint moved from pre-preview to pre-deploy so the preview is instant.
  - Replies never name the hosting provider: "our hosting", never "Vercel" (showing the `*.vercel.app` URL is fine).

  **2. The ibl.ai guidance now rides opencode's native instructions mechanism.** `guidance_with_identity` composes `IBLAI_INSTRUCTIONS` + per-user identity lines at spawn, and `write_iblai_guidance` writes it as `<config_home(session)>/opencode/AGENTS.md` — strictly before `cmd.spawn()`, with a failed write aborting the spawn (opencode silently ignores missing instruction files). opencode re-reads the file on every model call, so the guidance reaches main turns, subagents, ACP, and — for the first time — on-device (ollama/foundry) sessions, which bypass the loopback proxy entirely and previously ran with skills wired but zero guidance. The proxy's in-flight body injection is removed; the file never lands outside the per-session ibl.ai config home and is torn down with the session.

## Why

- The setup flow was auto-everything (scaffold, deploy, never localhost); users should consent to each step, once, conversationally.
- Proxy body injection only covered cloud sessions and existed outside opencode's own prompt state; the AGENTS.md convention is the mechanism opencode already re-reads per call.

## Proof & tests

- Empirically verified against the pinned opencode 1.18.13 binary: a scratch `XDG_CONFIG_HOME` with `opencode/AGENTS.md` + a stub chat/completions server capturing the request body — the guidance arrives in the system message (`Instructions from: …/AGENTS.md`). The recipe is recorded in `AGENTS.md` for future `OPENCODE_VERSION` bumps.
- Guard tests rewritten/extended: 3-step arc, ask-first preview, ask-once deploy, and never-name-the-hosting-provider pins; new unit tests for `guidance_with_identity` composition and the AGENTS.md write/rewrite/clear/loud-failure paths (including that the file stays inside the session config home); proxy injection tests removed with the code.
- e2e ledger: pending checkpoint `code-27` added for the 3-step flow (same tool-calling-model harness gap as code-08..10/15); journey coverage checker green, no regression.
- `cargo test --offline`: all tests pass in both binary targets.

## Docs

- Repo `AGENTS.md` (and its `CLAUDE.md` symlink) gains a "Code mode: ibl.ai instruction layers" section: the three layers, the write-before-spawn ordering contract, the retired delivery paths that must not come back, and the pin-bump proof recipe.


razasaddiqi (approved):
looks good

PR #478, opened 2026-09-03
feat(code-mode): retire the prompt fork, make AGENTS.md the authored prompt, and keep tool-call narration out of the reply

null-crafter:
## Checklist

## What

The Code agent was still wordy after moving the guidance into the per-session AGENTS.md. Two root causes, two fixes.

**1. The 99-line opencode prompt fork is gone.** `opencode_build_prompt.txt` is now a two-line stub: identity plus the silence rule (no text before or between tool calls, one reply at the end). Its only job is to *exist* — opencode resolves `agent.prompt ? [agent.prompt] : SystemPrompt.provider(model)`, so an empty prompt would silently restore the built-in per-model prompt, which mandates step-by-step narration (verified against the pinned 1.18.13 binary). The fork's verbosity vectors — the inherited Markdown-formatting license, the review-format protocol, "a clear explanation of outcomes", the channels section — are dropped outright. Its genuinely load-bearing rules moved into `IBLAI_INSTRUCTIONS`, rewritten tersely and without GPT-isms (`apply_patch`, `multi_tool_use.parallel`): implement-don't-propose autonomy, smallest-correct-change/ASCII editing, Glob/Grep + parallel reads, git safety (never revert others' changes, no `git reset --hard`, non-interactive git), and concrete opener bans. The guidance now opens with the silence contract and closes with a one-line reminder, and is written for **every** spawn — the skills-wired gate is removed, since with a stub prompt the guidance is the agent's entire authored behavior.

**2. A deterministic backstop for GPT-style preambles.** The remaining narration was genuine model text emitted before tool calls ("Let me check…"), which the SDK concatenates into the reply bubble; reasoning was never the issue (it already streams separately). On each new `tool_call`, `TurnState::take_narration` moves any pre-tool text into the reasoning section (`opencode:reasoning`) and retracts it from the bubble (empty `full_content`, which the SDK's replace-semantics `onToken` honors), so the visible reply is what follows the *last* tool call. A turn that ends on a tool call with no trailing text falls back to the model's only text rather than an empty bubble. No provider parameter exists for this on the openai-compatible path (`textVerbosity` is Copilot-only in opencode), so this is prompt salience + a structural guarantee.

## Tests

- `the_build_prompt_is_a_minimal_suppressor_stub`: non-blank (the falsy-restore trap), starts with the identity line, carries the silence rule, stays under 400 bytes, and never re-admits the upstream narration mandates.
- Guard test pins the moved working-discipline rules, and that the silence contract opens the guidance and the reminder closes it.
- `pre_tool_text_is_reclassified_as_narration`: reclassification, buffer restart, final-reply fallback, reset.
- The always-on pinned-binary e2e (`opencode_run_carries_the_agents_md_guidance_on_every_model_call`) still proves the real 1.18.13 binary carries the guidance on every agent turn under the stub.
- `cargo test --offline`: 123 passing per target. e2e ledger: pending checkpoint `code-28` (pre-tool narration lands in the thinking section), coverage checker green, no regression.

## Docs

Repo `AGENTS.md` (`CLAUDE.md` symlink): layer 1 is the two-line suppressor stub with the falsy-trap explanation and a "never grow it back into a fork" rule; layer 2 lists the moved rules and the unconditional write; the streaming backstop is documented alongside.


razasaddiqi (approved):
looks good

PR #505, opened 2026-09-10
Feat/mobile local llms and code

razasaddiqi:
## Checklist

### web-utils: platform gates for Tauri mobile
- Add `isTauriEmbeddedLLMHost()`: true on Tauri iOS and Android, where the host ships an embedded llama.cpp runtime behind Ollama's HTTP API.
- Add `isTauriCodeHost()`: true on any Tauri desktop, or on Tauri mobile when the host has set the `ibl_remote_code_ready` flag (phone paired with a desktop that proxies the opencode commands and events).
- Add `REMOTE_CODE_READY_KEY` constant for that pairing flag.
- Add `logToHost()`: fire-and-forget diagnostic that invokes the Tauri `log_fe` command so chat-path gate decisions show up in the `tauri ios dev` log stream. No-op outside Tauri and under test.
- Coding Mode stream, stop, and close now gate on `isTauriCodeHost()` instead of `isTauriDesktop()`. The not-available error now reads "Coding Mode needs the desktop app, or a phone paired with one."
- `useChat`: the local-LLM send path now also runs on Tauri mobile via `isTauriEmbeddedLLMHost()`. Coding Mode send, stop, and reset paths switch to `isTauriCodeHost()`.
- Export `isTauriDesktop` and `isTauriEmbeddedLLMHost` from the hooks index so web-containers can gate the model catalog. Streaming internals stay private.

### web-containers: per-device local model catalog
- `LocalModel` gains an optional `mobile: { size }` entry describing the phone-sized build. Phi-4 Mini, Llama 3.2, Qwen 3, and DeepSeek-R1 get mobile sizes.
- Add `localModelsForThisDevice()`: on Tauri mobile, only models with a mobile build are listed. Desktop and web see the full catalog.
- Add `localModelDisplaySize()`: returns the mobile build size on Tauri mobile, the desktop Ollama size elsewhere. Used for the row label, the too-large warning text, and the too-large capacity check.
- LLM provider modal and local-only provider cards in the Agent LLM tab now build from the per-device catalog, so phones don't get provider cards that open onto an empty picker.
- Agent LLM tab hides providers that have no server-side chat models and no on-device models for this device. The agent's active provider is always shown.
- `useModelDownload` passes the model id to the `check_disk_space` Tauri command so mobile hosts can budget the model's real size instead of a blanket requirement.

### web-containers: i18n for LLM tab copy
- `LLMProviderModal` and `LocalModelRow` now default their labels from the i18n catalog via `useT()` and `buildLLMTabLabels()` instead of the hardcoded English bundle. An explicit `labels` prop still overrides.
- `AGENT_LLM_TAB_LABELS` is documented as the English fallback and kept in sync with `en.json`, with a test enforcing string-for-string parity.
- Title-case copy updates in `en.json` and the fallback bundle: "Download Anyway", "Got It", "On-Device", "In Use", "Download Failed — Retry".

https://drive.google.com/drive/folders/1cTRIVh9M6Z-M_NVurTyemKcxYI0k1EvS?usp=sharing


null-crafter:
# PR 505 code review — `feat/mobile-local-llms-and-code`

**Scope:** issues attributable to razasaddiqi (Raza Fayyaz, `razasaddiqi@gmail.com`) only.
**Merge base:** `8cd77a99`. **Head:** `f61bd902`. **Reviewed:** 2026-09-11.

---

## Attribution

All thirteen findings are his. Nothing in this list belongs to another author.

| Check | Result |
| --- | --- |
| Commits between merge base and head | 5, all Raza Fayyaz |
| `local_llm.rs`, `remote_code.rs`, `remote_code_client.rs`, `app_update.rs` across all refs | only `c47715a1` and `255c7924`, both his |
| Advisory comments stripped from `pnpm-workspace.yaml` | present at merge base, gone in `c47715a1`, so not a merge artifact |

Two findings touch pre-existing code and are framed precisely below. Both remain his.

**Branch hygiene, separate from the findings:** main has advanced past the merge base.
`CHANGELOG.md`, `build-pr-test-image.yml` and `reusable-pr-docker-build.yml` show as
reversions in a two-dot diff. No finding touches them, but the branch needs main merged
again before it lands.

---

## Blockers

### 1. Committed yalc overrides break every install

`pnpm-workspace.yaml:13` — introduced by `c47715a1`, extended by `14a15736`.

Five `@iblai` packages are pinned to `file:.yalc/...`, and the lockfile resolves them as
directory entries. `.yalc/` is untracked and absent, so `pnpm install --frozen-lockfile`
cannot resolve them on CI or a clean clone.

Verified: no `.yalc` directory, `git ls-files .yalc` empty, `node_modules/` in this
checkout empty.

The same commit stripped the advisory rationale comments above `js-yaml`, `undici`,
`hono`, `dompurify`, `deepmerge-ts` and `nanoid` while keeping the version floors. The
merge base still carries them.

Also affected: `__tests__/web-utils-ios-local-llm-patch.test.ts` names
`patches/@iblai__web-utils@*.patch` as its fallback channel. That patch does not exist in
`patches/`, so the guard passes only while an uncommitted yalc build is linked.

**Fix:** restore registry versions and the advisory comments, or commit the patch file the
test names.

### 2. Desktop auto-update wired into the wrong entry point

`src-tauri/src/main.rs:2803` — omission in `255c7924`.

`app_update.rs`, `tauri_plugin_updater` and the two update commands were added only to
`lib.rs`. The desktop binary registers none of them.

The blamed line itself is old (`fc81473e`, initial commit), but Raza edited that exact
handler list in this PR to add four `remote_code` commands and the exit handler. He was
working in the desktop entry point and wired the updater only into the mobile one.

`AppUpdatePrompt` calls `invoke('check_app_update')` inside a catch that swallows the
"command not found" rejection, so the feature is silently dead on desktop while both
release workflows sign and publish the feed.

**Fix:** mirror `mod app_update`, the updater plugin and both commands into `main.rs`.

### 3. Mobile pairing capability never enabled

`src-tauri/tauri.conf.json:70` — from `255c7924`.

`app.security.capabilities` is an explicit allow-list and omits the new `mobile-pairing`.
Its five barcode-scanner permissions are never granted in a release build, so QR pairing
fails.

Debug builds mask this: the `debug_assertions` block at `lib.rs:2333` injects the same
permissions into the synthesized dev-origin capability.

**Fix:** add `mobile-pairing` to the allow-list.

### 4. Development URL baked into release builds

`src-tauri/build.rs:41` — from `c47715a1`.

Verified against the merge base with a positive control: `build.rs` there is eleven lines
and emits no `cargo:rustc-env` at all. Raza added resolution from `.env.local`,
`.env.production` and `.env`, emitted for every profile with no `debug_assertions` gate.

The reader it feeds, `get_app_url`'s `option_env!` at `lib.rs:111`, is pre-existing
(`24e675b2`, already on main). Before this change only a shell variable could reach it.
Now a dotenv file on the build machine does. Neither file exists in this checkout, so no
build is currently poisoned. The defect is the missing profile gate.

**Fix:** gate the emit on debug, or refuse to bake a non-HTTPS URL into a release.

---

## Correctness

### 5. Turn directory resolved before the session-id migration

`src-tauri/src/remote_code_client.rs:894`.

`run_turn` calls `ensure_directory` before `ensure_session`, which is what performs the
`new_chat_key` migration.

A chat with no mentor sends turn one under its ephemeral id, minting folder A. On turn two
the real id arrives, `ensure_directory` finds nothing, mints folder B and records it, and
the migration then skips because the key is already present. The turn prompts and
subscribes against B while reusing a server session that lives in A. Folder A is orphaned
with turn one's work.

Worse than a silent folder change: the two disagree for the rest of the chat. The guard
test at line 1507 asserts the no-mentor lookup returns `None`, which is the state that
triggers this.

**Fix:** run the migration before resolving the directory.

### 6. Base-name manifest match serves the wrong model

`src-tauri/src/local_llm.rs:151`.

`installed_model_path` matches whenever base names agree. With the smaller variant
installed, a pull of the larger one short-circuits as already installed at line 271 and
emits success without downloading. Chat then serves the smaller weights at line 407.

Both ids are distinct catalog entries and `resolve` at line 106 deliberately separates
them, so the two layers disagree.

**Fix:** apply the base-name fallback only when the request carries no tag.

### 7. Unpairing keeps the old machine's folders

`src-tauri/src/remote_code_client.rs:400`.

`remote_code_clear_host` clears url, password, mgmt, both alt lists and sessions, but
leaves `directories` and `mentor_directories`. Pair with a second desktop and
`resolve_directory` hands back an absolute path from the first, so the session is created
against a folder that does not exist there.

**Fix:** clear both maps alongside sessions.

### 8. Failed status probe counts as idle

`src-tauri/src/remote_code_client.rs:1014`.

The quiet-period probe folds a failed request into `None` via `.ok()`, then defaults to
`"idle"`. Two failures fifteen seconds apart, a Wi-Fi blip or a briefly sleeping desktop,
push the counter to the threshold and emit `ollama:done` with partial text as a complete
reply while the desktop keeps generating.

The event branch at line 1318 re-checks that content or tools exist. This branch has no
such check, so an empty reply can also be declared done.

**Fix:** treat a failed request as unknown, not idle.

### 9. Proxy secret leaks when enable fails

`src-tauri/src/remote_code.rs:499` — from `255c7924`.

The secret is registered before `apply_opencode_model`, `register_models_in_config`,
`ensure_workspace`, `free_port` and the companion bind, all of which return with `?` and
no unregister. It holds the user's platform token in the loopback proxy.

The abort paths at lines 549 and 632 do unregister, so the omission looks unintentional.
Each failed retry leaks another.

**Fix:** unregister on every early return after registration.

### 10. Dead host dropped without aborting the companion

`src-tauri/src/remote_code.rs:112`.

`status_locked` clears a crashed host but never aborts the companion task, so its server
keeps port-plus-one bound. Every other teardown path aborts it.

The next enable finds the port taken and falls back to a random one at line 573. A
hand-typed pairing that derives the management URL by convention
(`remote_code_client.rs:262`) no longer finds it. The same branch also skips the pid-file
removal and proxy unregister that `shutdown_sync` does.

**Fix:** reuse the `shutdown_sync` teardown in this branch.

### 11. UTF-8 reassembler discards the valid prefix

`src-tauri/src/local_llm.rs:868`.

When more than three trailing bytes are invalid, `push` clears the whole buffer and
returns `None`, throwing away already-valid text ahead of the bad bytes.

**Fix:** emit the valid prefix, drain only the invalid tail.

---

## Minor

### 12. Retry message contains two long space runs

`src-tauri/src/remote_code_client.rs:1019`. A single-line literal written as if it were
backslash-continued, so the user reads forty spaces mid-sentence. Verified with `cat -A`.

### 13. Changed JSON not formatted with prettier

`src-tauri/capabilities/default.json:5`, plus `mobile-pairing.json` and `tauri.conf.json`.
Arrays exploded, and the two capability files have no trailing newline. Contradicts the
repo's prettier rule. Prettier is not installed in this checkout, so this was verified
from the diff and a trailing-byte check.

---

## Verification

Nothing here is fixed. To confirm any fix:

- **Install:** `pnpm install --frozen-lockfile` must succeed on a clean clone with no
  `.yalc` present.
- **Rust:** `cd src-tauri && cargo test`. Findings 5, 6, 7, 8 and 11 each want a
  regression test that fails on the current code, per the repo's bug-fix rule.
- **Desktop updater:** a release build must answer `check_app_update` rather than
  rejecting with command not found.
- **Mobile pairing:** test a release build, not a debug one. Debug masks finding 3.
- **Formatting:** `pnpm prettier --write` on the three config files.

null-crafter (changes requested):
Please see the claude review above.

null-crafter:
# PR 505 code review — follow-up on `260ef530`

**Branch:** `feat/mobile-local-llms-and-code`
**Head:** `260ef530` "fix(tauri): adding fixes for the tauri issues ios android" (Raza Fayyaz, 2026-09-12)
**Previous review:** 13 findings at `f61bd902`
**Scope:** issues attributable to razasaddiqi only. All of them are.

---

## Part 1 — status of the previous thirteen findings

One commit answered the review. Twelve findings are closed, one is partially closed
by design. Each correctness fix landed with a regression test, as the repo rule
requires.

Verified directly in the tree, not taken from the report:

| Check | Result |
| --- | --- |
| `file:.yalc` refs in `pnpm-workspace.yaml` | 0 |
| `.yalc` refs in `pnpm-lock.yaml` | 0 |
| Stripped advisory comments (`GHSA-5p4m-2wfm-xmqj`) | restored |
| `app_update` in `main.rs` | present |
| `tauri_plugin_updater` in `main.rs` | present |
| `mobile-pairing` in `tauri.conf.json` capabilities | present |
| `ProbeVerdict`, `classify_probe` | `remote_code_client.rs` |
| `ProxySecretGuard`, `is_registered` | `remote_code.rs`, `opencode_proxy.rs` |
| `migrate_chat_key` | `remote_code_client.rs` |

| # | Finding | Status |
| --- | --- | --- |
| 1 | Committed yalc overrides | Fixed. Registry versions restored, advisory comments back, floors raised. |
| 2 | Updater wired only into `lib.rs` | Fixed. `main.rs` now has the module, the plugin and both commands. |
| 3 | `mobile-pairing` capability never enabled | Fixed. Added to the allow-list. |
| 4 | Dev URL baked into release | **Partial.** See Part 3. |
| 5 | Directory resolved before key migration | Fixed. `migrate_chat_key` extracted and called first, with a test. |
| 6 | Base-name manifest match | Fixed. Fallback only for untagged ids, with a test. |
| 7 | Unpairing keeps old folders | Fixed. Both directory maps cleared, with a test. |
| 8 | Failed probe counted as idle | Fixed. `ProbeVerdict::Unknown` leaves counters untouched, with a test. |
| 9 | Proxy secret leak on failed enable | Fixed. RAII `ProxySecretGuard`, with a test. |
| 10 | Dead host dropped without teardown | Fixed in code. **Its test is vacuous — see new issue 3.** |
| 11 | UTF-8 reassembler discards valid prefix | Fixed for the tail case. **Head case remains — see new issue 4.** |
| 12 | Long space runs in retry message | Fixed. |
| 13 | Unformatted JSON | Fixed. |

**Not independently verified:** the reviewing agent reports pulling
`@iblai/web-utils@2.4.5` from npm and running all four assertions of
`__tests__/web-utils-ios-local-llm-patch.test.ts` against it, concluding the missing
`patches/@iblai__web-utils@*.patch` is no longer needed. `node_modules` in this
checkout is empty, so that check happened elsewhere and is passed along rather than
confirmed.

---

## Part 2 — new issues introduced by `260ef530`

### 1. Empty `TAURI_APP_URL` disables the navigation allowlist — medium

`src-tauri/src/main.rs:2586`, new in this commit.

The fix for the white-window problem prepends the configured origin to the allowlist:

```rust
let app_origin = get_app_url();
let allowed = url_str.starts_with(app_origin.trim_end_matches('/')) || ...
```

`get_app_url()` at `main.rs:271` returns `std::env::var("TAURI_APP_URL")` verbatim, so
a bare `TAURI_APP_URL=` line yields `Ok("")`. `"".trim_end_matches('/')` is `""`, every
string starts with `""`, so `allowed` is unconditionally true and `if !allowed` never
fires. The desktop window will navigate anywhere.

The trigger is realistic. `dotenvy::from_path` loads `.env.local` then
`.env.production` at `main.rs:2174`, before anything reads env, and
`src-tauri/.env.example` documents the key at lines 15 to 19.

**Fix:**

```rust
let app_origin = get_app_url();
let app_origin = app_origin.trim_end_matches('/');
let allowed = (!app_origin.is_empty() && url_str.starts_with(app_origin)) || ...
```

**Also:** the hunk was not mirrored into `lib.rs`, whose desktop-shaped
`on_navigation` at line 2562 still carries the hardcoded list. `main.rs` is the shipped
desktop entry point so nothing is broken, but the repo's twin-entry-point rule wants
both guards to match.

### 2. The two `start()` tests collide on a process-wide `OnceLock` — medium

`src-tauri/src/local_llm.rs:1022` and `:1044`.

`start()` records the models dir with `let _ = MODELS_DIR.set(...)` at line 558. The
first call in the process wins; every later one is silently dropped. Two tests in the
same binary call `start()` with different dirs and no serialization:

- `embedded_server_answers_version_after_start` uses `ibl-live-<pid>/models`
- `models_dir_is_recorded_by_start_attempts_and_probeable` uses
  `ibl-models-dir-<pid>/models`, then asserts `models_dir()` equals its own dir

If the liveness test wins the race, that assertion fails. Order-dependent flake in
`cargo test`, which the pre-push hook runs.

**Fix:** serialize on a shared lock, reusing the pattern already in this repo at
`remote_code_client.rs:1428` (`static CONFIG_TEST_LOCK: tokio::sync::Mutex<()>`). Add
the equivalent to `local_llm.rs`'s `mod tests` and take it at the top of both tests.

### 3. The companion-abort assertion is vacuous — low

`src-tauri/src/remote_code.rs:912`.

`status_reaps_a_dead_child` builds the host with
`companion: tauri::async_runtime::spawn(async {})`, an immediately-completing future,
sleeps 200 ms at line 903, then asserts `companion.is_finished()`. That holds whether
or not `abort()` ran, so this half of finding 10 has no test that fails on the pre-fix
code. The secret-unregister half is properly pinned.

**Fix:** give the companion a future that never completes, so only the abort can
finish it:

```rust
companion: tauri::async_runtime::spawn(std::future::pending::<()>()),
```

### 4. UTF-8 reassembler still loses text when the bad bytes lead — low

`src-tauri/src/local_llm.rs:886`.

The tail case is fixed: a valid prefix followed by junk now emits the prefix. The head
case still drops everything. Feed `[0xF0, 0x9F]`, held as a plausible partial sequence,
then `b"hello world"`. `valid_up_to()` is 0 because the stale bytes sit at index 0,
`len - valid` exceeds 3, and the buffer is cleared with `None` returned. All eleven
good characters are lost.

Not a regression, and lower severity than the original, but the same class of loss.

**Fix:** when the invalid bytes lead, drop only the leading invalid sequence and retry
the remainder rather than clearing. Add the head case to
`utf8_stream_drops_invalid_bytes`.

### 5. `android-branding.sh` writes into a directory it never creates — low

`scripts/android-branding.sh:133`.

Under `set -euo pipefail` (line 13),
`cat > "$RES/../java/ai/ibl/mentorai/MainActivity.kt"` has no `mkdir -p`. The script's
only `mkdir` is at line 35, for `mipmap-anydpi-v26`. If the generated package path ever
differs, from an identifier change or a template that emits `kotlin/`, the redirect
fails and the script aborts after the icons were already copied, leaving a
half-branded project.

The real path could not be verified: `src-tauri/gen/android` is gitignored and absent
here. `src-tauri/gen` exists and contains only `apple`, so this is a meaningful check
rather than a missing-directory false negative.

Separately, the heredoc replaces the whole file with a class whose `onCreate` calls
only `super.onCreate`, so any other template content in that file is discarded.

**Fix:** `mkdir -p` the parent before the redirect, and fail with a clear message if
the expected package directory is missing rather than writing a file the build will
not pick up.

---

## Part 3 — open decision, not a defect

`src-tauri/build.rs:47` now gates the baked `TAURI_DEV_URL` on `PROFILE == "release"`
and drops non-https values with a `cargo:warning`. An **https** value still bakes into
a release build, which the comment states is deliberate so a hosted platform can be
targeted.

Given that iblai.org is the development-time platform, a release built on a machine
whose `.env.local` points at the org deployment ships aimed at dev. Confirm that is
intended before treating finding 4 as closed.

---

## Part 4 — still open from the original review

The branch is behind `origin/main`. `CHANGELOG.md`, `build-pr-test-image.yml`,
`reusable-pr-docker-build.yml` and the `package.json` version read as reversions in a
two-dot diff. Main needs merging again before this lands.

---

## Verification

- `cd src-tauri && cargo test` for new issues 2, 3 and 4. Run issue 2 several times,
  or alternate `--test-threads=1`, to confirm the flake is gone.
- New issue 3's test must fail on the pre-fix `status_locked`. Check by temporarily
  removing the `abort()` call; if it still passes, it is still vacuous.
- New issue 1 has no automatic coverage. Pinning it means extracting the allowlist
  predicate from the closure, which is the smallest honest way to test it.
- New issue 5 cannot be verified in this checkout. It needs a machine that has run
  `cargo tauri android init`.
- `pnpm prettier --write` on any touched `.ts` or `.json`, per the repo rule.

razasaddiqi:
### Code review

Found 1 issue:

1. `local_llm::start()` is not race-safe (bug in `src-tauri/src/local_llm.rs`: the "Idempotent" `server_url()` check at L609 is not synchronized with `SERVER_URL.set(url)` at L636 — two concurrent callers (e.g. app startup in `lib.rs` and `ensure_started` triggered by a model download) can both pass the check, both bind a `TcpListener`, spawn two `ibl-local-llm` threads, and only one of them wins the `OnceLock::set`. The losing thread stays alive on a port that is never returned to callers, so callers race against which URL they observe and one server thread is permanently unreachable.)

https://github.com/iblai/mentorai/blob/78a3310c1e73d9420613ae6ec0afea8e0bd3fa5a/src-tauri/src/local_llm.rs#L602-L638

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

razasaddiqi:
### Code review — follow-up on `75796ccc`

No issues found. Follow-up review of the delta since [my prior review](https://github.com/iblai/os/pull/505#issuecomment-5643713241) at `78a3310c` confirms the race in `local_llm::start()` is now fully synchronized: `START_LOCK` is acquired before `url_slot.get()` and held across bind → spawn → set, and the added `concurrent_starts_share_one_server` regression test (8 threads + `Barrier`, fresh `OnceLock`/`Mutex`) would have failed on the pre-fix code.

https://github.com/iblai/mentorai/blob/75796ccc853a1d17d152dd8c67131d9503ed6492/src-tauri/src/local_llm.rs#L602-L654

https://github.com/iblai/mentorai/blob/75796ccc853a1d17d152dd8c67131d9503ed6492/src-tauri/src/local_llm.rs#L1163-L1197

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

null-crafter:
# PR 505 code review — current state

**Branch:** `feat/mobile-local-llms-and-code`
**Head:** `82281852`
**Reviewed:** 2026-09-14
**Scope:** issues attributable to razasaddiqi's changes.

Fifteen open findings. **Seven block the merge**, eight do not, and one is a release
decision rather than a defect. Everything from the first three review rounds is closed.

Every finding was verified against the tree rather than relayed. One item reported as
high severity in the latest round is **refuted** and documented at the end so it does not
get re-filed.

---

## Blocking the merge

### B1. A real error is reported as the wrong thing

`src-tauri/src/remote_code_client.rs:1173`

The event gate drops anything without a session id:

```rust
let for_us = props.get("sessionID").and_then(|s| s.as_str())
    .map(|s| s == remote).unwrap_or(false);
if !for_us { continue; }
```

`session.error` is the one event opencode declares that field optional for. Verified at
the pinned version:

```
$ git -C ~/Projects/github/opencode show v1.18.13:packages/schema/src/v1/session.ts
652:  type: "session.error",
654:    sessionID: Schema.optional(SessionID),
```

**Scenario.** The desktop's model call fails unattributed, on auth or a provider outage.
The phone discards the event, streams nothing, and about thirty seconds later the idle
probe reports "The desktop went idle without producing a reply — try again". The real
cause never reaches the user.

**Fix.** Let `session.error` through the gate when the field is absent; keep the strict
check for `message.part.delta` and `permission.*`, where upstream marks it required.

### B2. A platform switch silently bricks phone turns

`src-tauri/src/remote_code.rs:538`

The already-running fast path re-registers the proxy upstream and token, then returns at
line 551, never reaching `apply_opencode_model` or `register_models_in_config`.

**Scenario.** The user switches platform or account and the frontend re-enables. The
upstream and token are corrected, but the serve keeps the previous user's identity lines
and only the previous platform's model ids. A turn asking for a model from the new
platform hits the async `ProviderModelNotFoundError` that the code's own comment at line
608 warns about: no event, no message, the turn spins to the thirty-minute deadline.

### B3. Chat scroll positions are computed from placeholder heights

`components/chat/chat-messages/index.tsx:140` and `:155`

Every message row now carries `[content-visibility:auto]` with
`[contain-intrinsic-size:auto_120px]`, and 240px for AI rows. Rows that have never
rendered report the placeholder height.

Three call sites do arithmetic on that:

| File | Line | Use |
| --- | --- | --- |
| `components/chat/index.tsx` | 785 | `scrollToBottom` target |
| `components/chat/index.tsx` | 797 | `isAtBottom` test |
| `components/chat/index.tsx` | 1108 | load-older restore |
| `components/chat/chat-messages/user-message-bubble.tsx` | 112 | reply-to `scrollIntoView({block:'center'})` |

**Scenario.** Opening a forty-message chat, `scrollHeight` is a large under-estimate, so
the view lands mid-conversation and grows under the reader. The load-older restore
under-counts the prepended block the same way. Tapping a reply-to quote in a freshly
opened conversation scrolls to an offset derived from placeholders and lands well away
from the target once the real rows render.

This is the most common interaction in the app, and the regression comes from this PR's
own optimisation.

### B4. A failed chunk load leaves an undismissible modal

`components/app-update-prompt.tsx:103`

```
103  setInstalling(true);
105  const { listen } = await import('@tauri-apps/api/event');   // outside try
106  const unlisten = await listen(...)                          // outside try
113  try {
```

`<AlertDialog open>` at line 134 has no `onOpenChange`, and all three footer buttons are
`disabled={installing}` at 167, 176 and 184.

**Scenario.** The event chunk 404s after a deploy — this repo ships `ChunkErrorRecovery`
precisely because that happens. The promise rejects unhandled, `installing` stays true,
every button is disabled, and Escape and outside-click do nothing. The app sits behind a
dead dialog until restart.

**Fix.** Move `setInstalling(true)` below the imports, or bring 105–112 inside the `try`.

### B5. `ensure_session` clobbers config written during its own await

`src-tauri/src/remote_code_client.rs:776`

```rust
let mut cfg = read_config();                                  // 776
let dir = ensure_directory(session_id, tenant, mentor).await;  // 777 — writes config
…
write_config(&cfg)?;                                           // 814 — stale snapshot
```

`ensure_directory` mints a workspace over the network and calls `record_directory`, which
does its own `read_config` / `write_config`. That write is then overwritten by the
snapshot taken before it.

**Scenario.** Any turn that mints a folder loses the `directories` or
`mentor_directories` entry recording it. A concurrent turn in another chat, or a folder
pick through `set_opencode_workspace`, is clobbered the same way, and that chat mints a
second server session on its next turn, losing the conversation.

Every accessor here is an unlocked full-file read-modify-write; the one spanning an await
is the dangerous one.

### B6. Re-pairing keeps the previous desktop's folders

`src-tauri/src/remote_code_client.rs:320`

`remote_code_set_host` writes `url`, `password`, `mgmt`, `alt_urls` and `alt_mgmt`, and
leaves `sessions`, `directories`, `mentor_directories` and `session_dirs` untouched.

This is the defect the `clear_host` fix closed, reachable through a different door: when
desktop A is off the network the popover renders `desktopUnreachable` **and** the scan
form, so scanning desktop B's QR re-pairs without ever calling `remote_code_clear_host`.
`resolve_directory` then returns A's absolute path, which does not exist on B, and
`ensure_session` scopes session creation to it.

**Fix.** Clear the four maps when the newly paired URL is neither the stored one nor among
`alt_urls`.

### B7. Desktop auto-restore fails silently and never retries

`components/chat-input-form/coding-mode-button.tsx:466`

The effect exists so that "every desktop restart silently bricks the paired phones" does
not happen — its own comment says so. It reads `tenant` and `dm_token` straight from
localStorage, and:

```js
      } catch {
        /* best-effort; the popover's Enable button remains */
      }
    })();
    …
  }, [mobile, sandboxed]);
```

**Scenario.** Either value is still empty when `remote_code_status` resolves. Rust rejects
with "sign in on this desktop before enabling phone access", the bare catch swallows it,
and the deps never change again. Phone access stays down for the whole session; the other
effect only re-enables when `st.running` is already true.

This is the silent-fallback shape the repo rules call out, in the one place written to
prevent the failure it now causes.

**Fix.** Retry on the auth-state change, or surface the failure.

---

## Non-blocking

### N1. `lib.rs` never reaps the remote-code serve child

`src-tauri/src/lib.rs:3157` ends at a bare `.run(tauri::generate_context!())` while
registering the four `remote_code` commands at 3114. `shutdown_sync` is called only from
`main.rs:2880`. Desktop ships `main.rs`, so no user impact today, but it violates the
repo's twin-entry-point rule.

### N2. The 300 s request timeout now bounds on-device inference

`src-tauri/src/lib.rs:1510`. On the new mobile path this bounds embedded llama.cpp
generation capped at `MAX_NEW_TOKENS = 1536`. A 3B model on an older phone at roughly
5 tok/s aborts mid-reply with "error decoding response body" after a partial answer was
shown. The non-streaming path at 1457 is tighter at 120 s.

### N3. `port + 1` overflows a `u16`, at two sites

- `src-tauri/src/remote_code.rs:657` — from `free_port()` or the persisted value.
- `src-tauri/src/remote_code_client.rs:279` — `companion_of` parses the port from a
  **hand-typed URL**, so user input reaches the overflow.

Panics in debug; wraps to 0 in release, binding a random port instead of the neighbour.
Use `checked_add(1)` and fall through.

### N4. Workspace line stale after pairing

`components/chat-input-form/coding-mode-button.tsx:583`. The refresh effect omits `mobile`
and the pairing state, so after a QR pair the connected view shows an empty workspace
while Select and New act on a folder the user was never shown. Clears on reopen.

### N5. Layout effect with no dependency array

`components/chat-input-form/use-overflow-fit.ts:124` runs after every render, calling
`getBoundingClientRect()` on each registered pill and forcing a synchronous reflow. Fires
on every keystroke across four to eight elements, on the phone webviews this PR is
otherwise optimising.

### N6. A lost enable race leaves the shared config holding a dead secret

`src-tauri/src/remote_code.rs:628`. If the post-provisioning re-check finds another enable
won, the function returns at 635 and `ProxySecretGuard` unregisters the secret — but
`apply_opencode_model` has already rewritten the shared `SERVE_CONFIG_KEY` `opencode.json`
with it. The running server re-reads that config and authenticates with a secret that no
longer exists, so every phone turn 401s until the next enable. Needs a race, but the
breakage persists.

### N7. Two `ollama:error` events per failure

`src-tauri/src/remote_code_client.rs:684`. `run_turn` emits `ollama:error` itself in three
places, then the command boundary re-emits for the same `generation_id`. Either return the
error silently or drop the re-emit.

### N8. The reaper misses the port actually rebound

`src-tauri/src/remote_code.rs:561` reaps `PREFERRED_PORT` and `PREFERRED_PORT + 1`, but the
port rebound is `persisted.port`. If an earlier enable fell back to an OS-assigned port and
the app was force-killed with no usable pid file, that orphan is never reaped, `free_port()`
returns yet another port, and every paired phone's stored `url` and `alt_urls` go stale
with no signal.

---

## Release decision, not a defect

`src-tauri/build.rs:47` gates the baked `TAURI_DEV_URL` on `PROFILE == "release"` and drops
non-https values with a warning. An **https** value still bakes in, deliberately per its
comment. Since iblai.org is the development-time platform, a release cut on a machine whose
`.env.local` points there ships aimed at dev. Merging is fine; cutting a release without
answering this is not.

---

## Reported and refuted — do not re-file

**"`prompt_async` is posted unscoped, so the phone sees no events."**
`src-tauri/src/remote_code_client.rs:1019`

The claim: `/event` and `/session/status` are scoped with `?directory=` but the prompt is
not, so the turn runs in the server cwd while the phone listens on the minted workspace,
and every event is filtered out.

The cited upstream files (`middleware/workspace-routing.ts`, `handlers/event.ts`) **do not
exist at the pinned `v1.18.13`**; that layout is from a newer release, and the local clone
is at `1.18.30`. Re-derived at the pin:

```
shared/workspace-routing.ts:24
  url.pathname.match(/^\/session\/([^/]+)(?:\/|$)/)?.[1]    ← matches /session/{id}/prompt_async
middleware/workspace-routing.ts:220-232
  const sessionID = getWorkspaceRouteSessionID(requestURL(request))
  const session   = sessionID ? yield* Session.Service.use(svc => svc.get(sessionID)) : undefined
  const plan      = yield* planRequest(request, session)
middleware/workspace-routing.ts:182
  directory: session?.directory || defaultDirectory(request, url)
```

The router extracts the session id from the path, loads the session, and prefers
`session.directory` over the `?directory=` ladder. An unscoped prompt lands on the
session's own instance.

The current code is precisely correct, not accidentally so. Line 21 of that same upstream
file is `if (url.pathname === "/session/status") return null`, and `/event` carries no
session id in its path. Those are exactly the two routes the Rust code scopes, and the only
two that need it.

---

## Closed since the earlier rounds

All eighteen findings from rounds one and two are resolved.

| Round | Item | Resolution |
| --- | --- | --- |
| 1 | Committed yalc overrides | Registry versions restored, advisory comments back |
| 1 | Updater wired only into `lib.rs` | Module, plugin and both commands now in `main.rs` |
| 1 | `mobile-pairing` capability disabled | Added to the allow-list |
| 1 | Directory resolved before key migration | `migrate_chat_key`, called first |
| 1 | Base-name manifest match | Fallback only for untagged ids |
| 1 | Unpairing kept old folders | Both directory maps cleared in `clear_host` |
| 1 | Failed probe counted as idle | `ProbeVerdict::Unknown` |
| 1 | Proxy secret leak on failed enable | RAII `ProxySecretGuard` |
| 1 | Dead host dropped without teardown | Full teardown in `status_locked` |
| 2 | Empty `TAURI_APP_URL` disabled the allowlist | New shared `nav_guard.rs` |
| 2 | `start()` tests raced on `MODELS_DIR` | `START_TEST_LOCK` |
| 2 | Vacuous companion-abort assertion | `spawn(std::future::pending::<()>())` |
| 2 | UTF-8 loss when bad bytes lead | `push` rewritten as an `error_len()` scan loop |
| 2 | Script wrote into an uncreated directory | `ACTIVITY_DIR` check that aborts loudly |

Two went past the ask. `nav_guard::under()` matches on an origin boundary, closing a hole
nobody reported: the old bare prefix test would have let
`https://os.ibl.ai.attacker.example` pass for `https://os.ibl.ai`. And the Android script
chose loud failure over `mkdir -p`, which is the better call.

---

## Provenance

The standing scope is razasaddiqi's changes. Not everything above is his code.

| Finding | Attribution |
| --- | --- |
| B1, B2, B4, B5, B6, B7, N3, N5, N6, N7, N8 | His own lines |
| B3 | The row wrappers are his (`255c7924`). Three of the four consumers are pre-existing: `chat/index.tsx` and `user-message-bubble.tsx:112` are both `fc81473e`. His change is what breaks them. |
| N1 | `lib.rs:3157` blames to `789b853a` (null-crafter). He added the exit handler to `main.rs` and the command registrations to `lib.rs` without the matching handler; the omission is his. |
| N2 | `lib.rs:1510` is from the initial commit. What is new is the on-device path routing through it, from `c47715a1`. |
| N4 | The effect line is `0bf6639b` (null-crafter). What is new is the `remoteHost` pairing state it fails to depend on, from `c47715a1`. |

---

## Also noted

- The new `fr` strings use straight apostrophes (`l'app`, `n'est`) where the surrounding
  catalog uses `’`. Harmless for ICU here since none precede a brace, but it breaks the
  repo convention. The four locales are otherwise in sync.
- The latest round verified opencode API shapes against the `v1.18.9` tag while the pin is
  `1.18.13`, and cited file paths from `1.18.30`. Its "checked and clean" list is weaker
  than it appears; treat it as unverified.

---

## Verification

- `cd src-tauri && cargo test` for the Rust half.
- B1 wants a test feeding a `session.error` frame with no `sessionID`, asserting the error
  surfaces instead of the idle message.
- B5 wants a test that a folder minted inside `ensure_directory` survives
  `ensure_session`'s write-back.
- B4 pins by rejecting the dynamic import in a vitest case and asserting the dialog closes.
- N3 pins at `port = 65535`, both sites.
- B3 and B7 have no honest unit test; verify in the running app.
- The turn tests use a fake server that ignores the query string, which is why scoping
  asymmetries do not show up there. Worth tightening if B5 lands near that code.
- `pnpm prettier --write` on any touched `.ts`, `.tsx` or `.json`, per the repo rule.

null-crafter (changes requested):
Pls check the claude review above and see if it makes sense.

null-crafter:
# PR 505 code review — round 6

**Branch:** `feat/mobile-local-llms-and-code`
**Head:** `c53681d1` — "fix(tauri): adding fix for the remote code multiple chat session" (razasaddiqi, 2026-09-15)
**Reviewed:** 2026-09-15
**Scope:** issues attributable to razasaddiqi's changes.

**No blocking findings remain.** The last blocker is fixed. Five non-blocking items and one
release decision carry over unchanged.

**CI is red — but the evidence says it is not this branch.** Detail below; please read that
section before treating the gate as a reason to hold the PR.

---

## What's left, at a glance

### Blocking (0)

None.

### Non-blocking (5)

- **Load-older scroll restore** — `chat/index.tsx:1108`. Opening a chat is fixed; only
  scroll-up-to-load-history still lands wrong. Still unverified in a running app.
- **Enable race leaves a dead secret in the shared config** — `remote_code.rs:689`. Needs
  two concurrent enables.
- **Layout effect with no dependency array** — `use-overflow-fit.ts:134`. Performance only.
- **Ambiguous prefix resolves silently** — `remote_code_client.rs`, `answer_for`.
- **Two desktop background emits bypass the marshalling helper** — pre-existing lines, not
  razasaddiqi's, desktop-only.

### Decision, not a defect (1)

- `build.rs:47` bakes an https `TAURI_DEV_URL` into release builds. Blocks cutting a
  release, not merging.

---

## The blocker is fixed

`c53681d1` reworks `ensure_session` in `src-tauri/src/remote_code_client.rs`:

- the session-reuse check is now a **scoped read-only block**, so no snapshot is held
  across an await;
- config is re-read **after** the session-create round-trip, and only this chat's two
  entries (`sessions`, `session_dirs`) are merged in;
- the comment names the failure directly — a config write landing in that window "was
  silently undone — that chat's next turn found no session and minted a second one, and
  its conversation was gone".

The regression test `a_session_created_during_another_chats_create_request_is_kept` is
**deterministic, not timing-based**: a `tokio::sync::Notify` gate parks chat A's create
request until chat B has written its own entries, so the race window is hit on every run
rather than when the scheduler happens to cooperate.

---

## The red CI is not this branch

Two consecutive failing runs at `c53681d1` (7 failures each, `new=7 known=0`), after a
green run at `82281852`. Three things rule the branch out:

**1. The commit cannot break a web suite.** `c53681d1` touches exactly one file,
`src-tauri/src/remote_code_client.rs`. It is Rust; the e2e suite is the web app.

**2. An unrelated PR fails the identical tests.** PR **os#467** fails the same four
Journey 52 cases:

```
e2e/journeys/52-tool-call-indicator-and-reasoning.spec.ts:20:7
e2e/journeys/52-tool-call-indicator-and-reasoning.spec.ts:102:7
e2e/journeys/52-tool-call-indicator-and-reasoning.spec.ts:185:7
e2e/journeys/52-tool-call-indicator-and-reasoning.spec.ts:490:7
```

os#505 run [`34900892033`](https://github.com/iblai/iblai-deploy-ops/actions/runs/34900892033)
versus os#467 run [`34958260222`](https://github.com/iblai/iblai-deploy-ops/actions/runs/34958260222).

**3. The platform moved underneath the runs.** The green run tested against prod release
`1.209.0`; both red runs against `1.221.0`.

### The one failure that looked branch-specific is not

`12-chat-sharing.spec.ts:272:9` (`sh-09`, message download) was the obvious suspect, since
this branch put `content-visibility: auto` on message rows and that test reads message
content. Both halves of that theory are wrong:

- `components/chat/ai-message-download.tsx` is **Conrad Mugabe's `074ca651`, already on
  `origin/main`** — it reached this branch through the main merge, not razasaddiqi's work.
- It builds the transcript from the `messages` array via
  `buildTranscript`/`buildMessageTranscript` (`components/chat/chat-transcript.ts`) — pure
  state, **no DOM read** — so `content-visibility` cannot affect it.

### Limit of this analysis

The CI log extracts are incomplete: **5 of 7** failing test names for os#505 and **6 of 12**
for os#467. Two of this PR's failures remain unidentified, so this is not a full set
comparison. What is established is that the largest cluster is shared with an unrelated
branch, and that the newest commit cannot have caused any web failure.

**Suggested next step:** re-run e2e, or compare against a main run at `1.221.0`, before
treating the gate as this PR's problem.

---

## Non-blocking detail

**Load-older scroll restore.** `chat-messages/index.tsx` defers `content-visibility`:
`skipOffscreen` starts false, every row renders for real, and skipping switches on two
`requestAnimationFrame`s later, after the initial scroll lands. That fixed
scroll-to-bottom on open and the reply-to jump. The effect resets only on
`[sessionId, hasMessages]`, and loading older messages changes neither, so
`chat/index.tsx:1108` still computes `scrollHeight - prevScrollHeightRef.current` while the
prepended rows have never been laid out. Browser-behaviour reasoning, not something I ran.

**Enable race.** `remote_code.rs:689` — `apply_opencode_model` still runs before the race
re-check, so a losing enable unregisters a secret the shared `SERVE_CONFIG_KEY` config
already names, and the running server 401s every phone turn until the next enable.

**Layout effect.** `use-overflow-fit.ts:134` still closes `});` with no dependency array,
measuring every registered pill after every render.

**Ambiguous prefix.** `answer_for` documents selection by "an unambiguous prefix" but takes
the first match, so options "Deploy now" and "Deploy later" resolve silently on `deploy`.
Falling through to the verbatim text would let the agent re-ask.

**Background emits.** `opencode_proxy.rs` and `opencode_installer.rs` call `app.emit(...)`
directly from an axum handler and an install task — the pattern `main.rs::emit_on_main`
documents as causing the "Not Responding" freeze. Desktop-gated, pre-existing.

---

## The `question` tool hang — fixed (round 5, unchanged)

Raised directly: *the remote code agent tried to use a Question tool which didn't exist in
OS.*

**Root cause** (`opencode_acp.rs`): the `question` tool parks a session until a client
answers over `/question/{id}/reply`. Under ACP opencode already withholds it, because the
tool is gated to the app, cli and desktop clients. **The phone talks to `opencode serve`,
where it was still on**, so setup questions parked every phone turn. The guard test names
the symptom: "the phone showed the question and then sat on the stop button forever."

**The fix, four layers:**

1. `enforce_permission_policy` writes `{"*": "ask", "question": "deny"}` on every spawn.
2. Written as a **permission**, not the `tools: {question: false}` shorthand — opencode
   turns that shorthand into a permission rule which a top-level string policy then
   overwrites. Recorded as verified against `1.18.13`: with the shorthand the model still
   got the tool.
3. Per-agent `permission` blocks are dropped, since one would take precedence and silently
   reopen it.
4. `IBLAI_INSTRUCTIONS` now says: "Ask each of these questions as plain text in your reply
   and end your turn there … There is no question tool: never try to call one."

**The phone path is covered:** `remote_code.rs` calls
`apply_opencode_model(SERVE_CONFIG_KEY, …)` — the same config writer ACP spawns use — and
`apply_opencode_model` calls `enforce_permission_policy` immediately before the process
starts.

**Safety net:** if a question arrives anyway, `pending_question` renders the options as a
numbered list, the turn ends, and the next message answers it via `answer_for` (see the
ambiguous-prefix item above).

**Upgrade residual, not a defect:** a `serve` already running keeps the config it started
with, since a running serve never re-reads `opencode.json`. Resolves on the next app
restart.

---

## Reported and refuted — do not re-file

**"`prompt_async` is posted unscoped, so the phone sees no events."**

Reported as high severity in round four. The cited upstream files
(`middleware/workspace-routing.ts`, `handlers/event.ts`) **do not exist at the pinned
`v1.18.13`**; that layout is from a newer release. Re-derived at the pin:

```
shared/workspace-routing.ts:24
  url.pathname.match(/^\/session\/([^/]+)(?:\/|$)/)?.[1]    ← matches /session/{id}/prompt_async
middleware/workspace-routing.ts:220-232
  const sessionID = getWorkspaceRouteSessionID(requestURL(request))
  const session   = sessionID ? yield* Session.Service.use(svc => svc.get(sessionID)) : undefined
  const plan      = yield* planRequest(request, session)
middleware/workspace-routing.ts:182
  directory: session?.directory || defaultDirectory(request, url)
```

The router extracts the session id from the path and prefers `session.directory` over the
`?directory=` ladder, so an unscoped prompt lands on the session's own instance. The
current code is precisely correct: that upstream file returns null for `/session/status`,
and `/event` carries no session id — exactly the two routes the Rust scopes, and the only
two that need it.

---

## Closed across rounds 1–6

All twenty-one findings from the earlier rounds are resolved, including: the committed
yalc overrides, the updater wired into the wrong entry point, the disabled
`mobile-pairing` capability, the base-name model match, the proxy-secret leak, the dead
host teardown, the empty `TAURI_APP_URL` allowlist hole (fixed with a shared `nav_guard.rs`
whose `under()` also closed an origin-boundary hole nobody reported), the UTF-8
reassembler, the dropped `session.error`, the stale served config after a platform switch,
the undismissible update modal, the re-pair folder leak, the silently-bricked auto-restore,
the `lib.rs` exit handler, the inference timeout, the port overflow, the double
`ollama:error`, the reaper scope, and now the multi-chat config clobber.

Two went past the ask, and one fix was for a bug no review raised: an **iOS main-thread
deadlock**, where `emit_on_main` now marshals every background emit, citing watchdog kills
(`0x8BADF00D`) during phone Code turns.

Also from the thread: razasaddiqi's own two automated reviews found and then confirmed the
fix for a race in `local_llm::start()`, where the idempotent `server_url()` check was not
synchronized with `SERVER_URL.set`; `75796ccc` holds `START_LOCK` across bind → spawn → set
with an eight-thread barrier test.

---

## Verification

- `cd src-tauri && cargo test` — confirm `a_session_created_during_another_chats_create_request_is_kept`
  fails on the pre-fix `ensure_session`.
- Re-run e2e, or compare against a main run at platform `1.221.0`, before attributing the
  red gate to this branch.
- The load-older residual needs the running app: open a long chat and scroll up past the
  load-older threshold.
- The ambiguous-prefix item pins as a unit test: two options sharing a prefix, assert the
  typed prefix passes through verbatim.
- `pnpm prettier --write` on any touched `.ts`, `.tsx` or `.json`, per the repo rule.

null-crafter (approved):

PR #511, opened 2026-09-14
chore(auth): expose NEXT_PUBLIC_IBL_PER_TAB_AUTH in runtime env.js

sonegillis1:
## What
`entrypoint.sh` generates `window.__ENV__` at container start (the runtime env the app reads via `config`). This adds the per-tab auth flag so it can be configured per-environment at runtime:

```
NEXT_PUBLIC_IBL_PER_TAB_AUTH: "${NEXT_PUBLIC_IBL_PER_TAB_AUTH}",
```

## Why
The per-tab auth feature reads `NEXT_PUBLIC_IBL_PER_TAB_AUTH` from `window.__ENV__` first, then the build-time value. Without this line the flag can't be flipped at runtime in mentorai deploys (skillsai already exposes it the same way).

## Safety
No-op until `NEXT_PUBLIC_IBL_PER_TAB_AUTH` is set in the environment **and** the per-tab feature code is present — an unset var just emits an empty string. One-line, `entrypoint.sh` only.


sonegillis1:
Superseded — moved this change onto the per-tab PR #467 instead of a standalone PR.

PR #513, opened 2026-09-15
fix(tauri): adding fix for icons, minimum os version & os platform for model id

razasaddiqi:
## Checklist

### Fixed
- **Code (phone): switching chats mid-turn no longer loses a chat's session.** `ensure_session` snapshotted the session map before the create request and wrote the whole snapshot back after it, so a session recorded by another chat during that request was overwritten and that chat silently started over. The write-back now re-reads the file after the request and merges only the current chat's entries. Regression test parks chat A's create request, writes chat B's session in the window, and asserts B survives (fails on the old code). — round-5 blocking item P2
- **Release builds no longer read the repo's `.env.local`.** The desktop loaded `src-tauri/.env.local`/`.env.production` through a path compiled in at build time — in release builds too — so a DMG built on a dev machine (a) opened the developer's localhost instead of production and hung on "Loading…", and (b) took the dev platform domain (`iblai.org`), asked it for the tenant's model list with a production token (401) and registered only the default model, so every phone Code turn failed with `Model not found: iblai/<model>`. The dotenv loads (`main.rs`, `lib.rs`) and `opencode_proxy::local_env` are now debug-only; release resolution is shell env → bundled `.env` → production defaults, identical to a CI-built DMG. Test `release_builds_ignore_the_checkout_dotenv` runs in the release profile.
- **iOS `Product → Archive` in Xcode works.** Tauri's stock `xcode-script` build phase only works when `cargo tauri ios build` is the parent process (it fetches options over a local socket), so archiving from Xcode failed with `PhaseScriptExecution … Connection refused` (after first failing with `cargo: command not found` on Xcode's bare `PATH`). The "Build Rust Code" phase now runs `gen/apple/build-rust.sh`, which builds the crate's static library for each requested arch with the iOS SDK flags and copies it to `Externals/<arch>/<config>/libapp.a`, with cargo and Homebrew (`cmake`, for llama.cpp) on `PATH`. `cargo tauri ios build --export-method app-store-connect` still works.
- **App Store Connect validation.** All four rejections fixed: app icons had an alpha channel (flattened, all 18); `libapp.a` was copied into the bundle via a stray Resources phase entry (now link-only); the iOS version (`0.95.19`) was below the last approved `1.1.19`; `MinimumOSVersion` was 14.0 (Apple requires ≥ 15.0 from spring 2027) — deployment target raised to 15.0 in `tauri.conf.json`, the Xcode project and `project.yml`.

### Changed
- **All mobile builds default to `https://os.ibl.ai`**, the same production host as desktop (was `mentorai.iblai.app` on iOS/Android; a stale LAN address for debug builds with nothing baked in). Universal links accept both `os.ibl.ai` and `mentorai.iblai.app`. `scripts/android-release.sh` and `.env.example` updated.
- **iOS versioning has its own line**, via Tauri's per-platform overlay `src-tauri/tauri.ios.conf.json`: `version` (marketing, currently `1.1.20`) and `bundle.iOS.bundleVersion` (plain build counter, currently `2`). Bump the build number per upload, the version per release; `tauri.conf.json`'s `version` stays the desktop/updater line.


null-crafter:
# PR 513 code review

**Head:** `52b99db7` — "fix(tauri): adding fix for icons, minimum os version & os platform for model id"
**Base:** `main` · **Reviewed:** 2026-09-16

Twenty-five files, but nineteen are binary icon replacements — the reviewable surface is
seven. The stated purpose lands, and the runtime half of the "dev machine config leaks into
a release build" class is fixed properly in both entry points.

One blocking item, two worth fixing, three nits.

---

## Blocking

### B1. The Xcode project now runs a script the repo does not contain

`src-tauri/gen/apple/ibl-ai-os.xcodeproj/project.pbxproj` replaces the self-contained build
phase:

```diff
-shellScript = "cargo tauri ios xcode-script -v --platform ${PLATFORM_DISPLAY_NAME:?} … ${ARCHS:?}";
+shellScript = "\"$SRCROOT/build-rust.sh\"";
```

`build-rust.sh` is not in the repository. Listing the directory at this head — rather than
trusting a 404 — returns exactly two entries:

```
$ gh api repos/iblai/os/contents/src-tauri/gen/apple?ref=52b99db7
dir  Assets.xcassets
dir  ibl-ai-os.xcodeproj
```

`src-tauri/gen/apple/` is gitignored (`.gitignore:60`), so those two were force-added and
the new script was not — a plain `git add` would skip it silently. It is absent from a
fresh checkout too.

There is no iOS workflow under `.github/workflows/`, so this will never show up as a red
gate. It shows up as **iOS building only on the author's machine**: any clone, including a
fresh one by the author, fails at the Build Rust Code phase with "no such file". The
command it replaced worked from a clone with no extra files.

Either commit `build-rust.sh` with `git add -f`, or keep the self-contained
`cargo tauri ios xcode-script` invocation.

---

## Should fix

### B2. Universal links on `os.ibl.ai` cannot fire

The deep-link handler gains a branch for the new host:

```rust
// Both production hosts are in the associated-domains entitlement.
let is_universal_link =
    scheme == "https" && (host == "os.ibl.ai" || host == "mentorai.iblai.app");
```

A universal link needs two halves: the entitlement **and** an
`apple-app-site-association` file served by the host. The served half is missing, with the
old host as a positive control:

```
https://mentorai.iblai.app/.well-known/apple-app-site-association
  → HTTP 200, application/json
    { "applinks": { "details": [ { "appID": "L4FWRM8W5Z.ai.ibl.mentorai",
                                   "paths": ["/sso-login", "/sso-login/*", "/mobile/*", "/test/*"] } ] } }

https://os.ibl.ai/.well-known/apple-app-site-association
  → the Next.js HTML 404 page
```

So iOS never registers `os.ibl.ai` for the app, and the new branch is unreachable for it.
Nothing tracked carries `applinks` either — `entitlements.mac.plist` and
`entitlements.devid.plist` are macOS and have none — and `scripts/verify-entitlements.sh`
still references only `mentorai.iblai.app` throughout.

This is a deployment gap the PR surfaces rather than causes: since the same commit moves
the mobile default host to `os.ibl.ai`, the association file needs to be served there
before the branch does anything. The custom scheme (`iblai-mentor://`) is unaffected.

### B3. The release-leak fix is complete at runtime, not at compile time

The runtime half is good: dotenv loading is now `#[cfg(debug_assertions)]` in **both**
`main.rs` and `lib.rs`, and `local_env` returns `None` in release, with a test.

The compile-time path is still open. `build.rs` bakes `TAURI_DEV_URL` from
`.env.local` / `.env.production` into **release** builds whenever the value is https — it
only rejects non-https — and `get_app_url` reads it before the new default:

```rust
// Mobile: compile-time env var (set during build)
if let Some(url) = option_env!("TAURI_DEV_URL") {
    return url.to_string();
}
#[cfg(any(target_os = "ios", target_os = "android"))]
return "https://os.ibl.ai".to_string();
```

A release built on a dev machine whose dotenv points at the org platform therefore still
opens that host — the same class of bug this commit fixes at runtime, left open one layer
down. It also makes the new comment inaccurate: "Release builds behave the same everywhere:
shell env, then the bundled/cwd `.env` below, then the compiled-in default" skips the
compiled-in `TAURI_DEV_URL`, which wins over all of them.

---

## Nits

**N1.** `tauri.ios.conf.json` pins `bundleVersion: "2"`. App Store Connect rejects a
duplicate build number for the same version, so the next upload needs a manual edit. Worth
deriving it or documenting the bump.

**N2.** The same file pins `version: "1.1.20"` while `tauri.conf.json` is `0.95.19`.
Presumably deliberate — App Store versions can't go backwards — but it is now a second
place to maintain, so worth confirming it's intended rather than a copy-paste.

**N3.** `release_builds_ignore_the_checkout_dotenv` can fail on a developer machine. The
debug branch asserts:

```rust
assert!(value.is_some() || !has_local, "debug builds read the checkout dotenv");
```

`has_local` is true when either dotenv file *exists*, but `value` is `Some` only when one
contains `IBLAI_PLATFORM_DOMAIN`. A `.env.local` without that key fails the assertion.
Test for the key, not the file.

---

## Good

- The dotenv gating went into **both** entry points, respecting the twin rule rather than
  only the one being debugged.
- The hardcoded `192.168.1.46:3001` mobile debug fallback is gone — that was a landmine for
  anyone but its author.
- `local_env`'s release gate ships with a test rather than just a comment.
- `libapp.a` dropped from the Resources build phase, where a static library did not belong.
- The iOS deployment target 14 → 15 matches the new `minimumSystemVersion: "15.0"` rather
  than drifting from it.

---

## Verification

- **B1:** `gh api repos/iblai/os/contents/src-tauri/gen/apple?ref=52b99db7` lists the
  directory; a fresh clone plus an iOS build reproduces the failure.
- **B2:** `curl https://os.ibl.ai/.well-known/apple-app-site-association` — expect JSON,
  currently HTML.
- **B3:** build a release with `TAURI_DEV_URL=https://…` in `src-tauri/.env.local` and
  confirm the binary opens that host instead of `os.ibl.ai`.

---

## Verdict

**Changes requested** — on B1 alone: the tracked Xcode project references a file the
repository does not contain, so iOS builds only where that untracked file already exists.

B2 is a deployment gap this PR surfaces rather than introduces, and B3 is a pre-existing
path left open while a new comment claims it is closed; neither needs to block if they are
tracked separately.

This is an assistant review and does not replace the external reviewer's.

null-crafter:
# PR 513 code review — round 2

**Head:** `c5d53f9c` (`037ee3a5` "fix for code review and ios permissions", plus a merge of main)
**Reviewed:** 2026-09-16 · follow-up to [the first round](https://github.com/iblai/os/pull/513#issuecomment-5687350719)

Two of the three findings are properly fixed, with tests. One is unchanged. The new commit
adds one issue of its own.

---

## Fixed

**B1 — the missing build script.** `src-tauri/gen/apple/build-rust.sh` is now present and
executable (mode `100755`, matching `scripts/verify-entitlements.sh` as a control), along
with `project.yml` and `ibl-ai-os_iOS/`.

The header also explains the cause better than my report did: the stock
`cargo tauri ios xcode-script` only works when `cargo tauri ios build/dev` is the parent
process, because it fetches its options from the CLI over a local socket — so
Product > Archive failed with "Connection refused". Replacing it with a self-contained
script is the right call, not just a workaround.

**B3 — the compile-time half of the dev-URL leak.** Closed at the right place:

```rust
fn dev_url_from_env_files(release: bool) -> Option<String> {
    if let Ok(v) = std::env::var("TAURI_DEV_URL") { return Some(v); }
    if release { return None; }          // ← dotenv never reached in release
    …
}
```

The doc names it exactly — "the compile-time twin of the runtime leak `local_env` / the
dotenv loads used to have". Three tests cover it, including
`only_https_may_be_baked_into_a_release` and `a_release_build_ignores_the_checkout_dotenv`,
with a static `ENV_LOCK` serialising the env-var mutation between them so they can't race.

**N3 — the flaky dotenv test.** Now keyed on whether a file actually *sets* the key, with
`assert_eq!(value.is_some(), sets_key)` — tighter than the fix I suggested.

---

## Still open

**B2 — universal links on `os.ibl.ai`.** Rechecked today, unchanged:

```
https://os.ibl.ai/.well-known/apple-app-site-association          → HTTP 404, text/html
https://mentorai.iblai.app/.well-known/apple-app-site-association → HTTP 200, application/json
                                                                    (positive control)
```

The `host == "os.ibl.ai"` branch stays unreachable until that file is served. A deployment
task rather than a code one, but it needs to land before the new default host's deep links
are relied on.

**N1 —** `tauri.ios.conf.json` still pins `bundleVersion: "2"`; the next App Store upload
needs a manual bump.
**N2 —** it still pins `version: "1.1.20"` against `tauri.conf.json`'s `0.95.19`. Fine if
deliberate, worth confirming.

---

## New this round

### S1 (should fix) — `project.yml` still carries the command it replaced

This commit adds `src-tauri/gen/apple/project.yml`, the XcodeGen spec that
`project.pbxproj` is generated from. The two now disagree:

```
project.pbxproj   shellScript = "\"$SRCROOT/build-rust.sh\"";
project.yml       script: cargo tauri ios xcode-script -v --platform … ${ARCHS:?}
```

Regenerating from the spec restores exactly the command whose failure motivated
`build-rust.sh` — per that script's own header, Archive dies with "Connection refused".

Worth raising because the same reasoning is already applied one file over.
`src-tauri/Info.plist` carries the comment: *"Merged over gen/apple's Info.plist on every
`cargo tauri ios build`, so a regenerated Xcode project keeps them."* That awareness of
regeneration just didn't reach `project.yml`.

**One thing I could not determine:** whether `cargo tauri ios build` re-runs XcodeGen from
`project.yml`, or only `cargo tauri ios init` does. If it regenerates on build, the B1 fix
is already inert and this is blocking rather than latent. Worth settling before depending
on the fix.

### S2 (nit) — a doc reference to a script that isn't here

`build.rs`'s new comment points at `scripts/android-release.sh`, which is not in the repo.
Control: 27 files listed under `scripts/`, with `android-branding.sh`, `android-dev.sh` and
`android-add-deep-link-scheme.sh` present. The sibling reference,
`gen/apple/build-rust.sh`, does exist.

---

## Checked and clear

Three things I suspected and confirmed are fine, recorded so they aren't re-litigated:

- **`Externals/<CONFIGURATION>` casing.** `build-rust.sh` writes lowercase
  `release`/`debug`, and I expected that to miss Xcode's `$(CONFIGURATION)` in the library
  search paths. It doesn't: `project.yml` declares `configs: {debug: debug, release:
  release}`, so the configuration names are lowercase and the paths line up.
- **`--no-default-features`.** Safe. `[features]` has only `embedded-llm` and no `default`
  key, and the iOS inference engine is `cfg(target_os)`-gated rather than feature-gated, so
  the flag changes nothing on iOS.
- **Two Info.plists.** Deliberate, not a duplicate: `src-tauri/Info.plist` is the source
  Tauri merges over the generated one, and both carry the four usage strings.

The iOS permission strings themselves read well — each says what the app does with the
resource, which is what review looks for.

---

## Verification

- **S1:** `rg -n 'script:' src-tauri/gen/apple/project.yml` against
  `rg -n 'shellScript' src-tauri/gen/apple/ibl-ai-os.xcodeproj/project.pbxproj`.
- **B2:** `curl https://os.ibl.ai/.well-known/apple-app-site-association` — expect JSON.
- **B1/B3:** `cd src-tauri && cargo test` covers the three new `build.rs` tests; an Xcode
  Archive exercises the script end to end.

---

## Verdict

**Approved**, with caveats. The two blocking items from the first round are fixed, each
with a test that would fail on the pre-fix code, and nothing in the current artifact is
broken.

Caveats — none a reason to withhold approval:

- **B2** must land before `os.ibl.ai` deep links are relied on.
- **S1** should be aligned before anyone regenerates the Xcode project, and becomes
  blocking if `cargo tauri ios build` turns out to regenerate.
- **N1** will fail the next App Store upload.

This is an assistant review and does not replace the external reviewer's.

null-crafter (approved):

PR #515, opened 2026-09-16
fix(tauri): adding fix for the ios code and UI responsive

razasaddiqi:
## Checklist

### Summary
Four user-visible fixes on the phone app, plus the plumbing they needed:

1. **QR pairing works on the first scan.** iOS raises its Local Network prompt on the first socket an app opens toward the LAN and fails that socket immediately; our pairing was a single probe with no retry, so the first scan always lost the race and showed "connection failed".
2. **A Code turn survives the lock screen / app switcher.** iOS reclaims a suspended app's sockets; the SSE stream died and the turn failed with "event stream broke" even though the desktop was still working.
3. **Local Network permission is asked at first open and handled end to end** (denied → the Code panel explains and offers Open Settings; the switch is re-read live in both directions).
4. **Confirmation dialogs are phone-safe** (approval mode, update prompt, Cowork consent, delete modals…): no more edge-to-edge/square rendering below the `sm` breakpoint.

Also: a turn abandoned by a force-quit is stopped on the desktop at the next launch, and the iOS build number moves to 5.

### What changed

**Pairing** — `src-tauri/src/remote_code_client.rs`
- `remote_code_set_host` now races every address the QR advertises (`first_answering`) with a 4 s cap per probe and keeps probing every 0.75 s for up to 25 s — long enough to read the prompt and tap Allow. A desktop that *answers* and refuses (401) fails immediately; only network-level failures retry. An empty candidate list errors instead of spinning.
- Panel makes ONE host call carrying all addresses (looping per address would multiply the wait) and shows "If your phone asks to find devices on your local network, tap Allow." while connecting.

**Turn stream** — `remote_code_client.rs`, `lib.rs`
- `open_event_stream` / `reconnect_event_stream` (backoff, 2 min budget). The loop reconnects on stream error/end, on 30 s without opencode's 10 s `server.heartbeat`, and immediately on foreground (`WindowEvent::Resumed` → `app_resumed()`, mobile only). The reconnect check runs before the turn deadline so a turn that finished during a long lock completes instead of timing out.
- `catch_up` reconciles after a reconnect: re-reads the reply from `/session/{id}/message?limit=N` (widening 16→64→256→all until it reaches the user message), emits only what the phone hasn't shown (`TokenCoalescer::catch_up_to`; the frontend renders `full_content`), re-emits tool parts, re-announces pending `/permission` entries (auto mode answers them), ends the turn on a pending `/question`, and emits `ollama:done` when `/session/status` says idle.

**Force-quit** — `HostConfig.in_flight` (server session → dir) is held by an `InFlight` guard for the turn's lifetime; `abort_abandoned_turns()` runs at mobile startup and posts `/session/{id}/abort` for entries still busy on the desktop, forgetting only what it handled. The desktop's Phone Access server is never touched.

**Local Network permission** — new `src-tauri/src/local_network.rs`
- iOS has no API for this permission. The probe is the documented behaviour itself: one empty UDP datagram to the Wi-Fi subnet's `.1` neighbour (`en0` via `getifaddrs`). Sent → granted; `EHOSTUNREACH` → refused. Telling "prompt still up" from "Don't Allow": a system alert makes the app inactive (`UIApplicationWillResignActive`/`DidBecomeActive` observers), so a refusal while inactive waits, a refusal while active is the answer (after a 1 s grace for the alert to appear on the first ask; instantly once the persisted `local_network_asked` marker exists). Nothing is cached: the Settings switch changes it live both ways.
- Asked at launch (2 s after setup, 90 s window, off the main thread) and from the panel (`local_network_status`, 20 s window). `open_app_settings` opens `app-settings:`. Both commands added to `permissions/default.toml` + `capabilities/default.json`. Android reports granted; desktop carries the module unused.
- Code panel: `denied` replaces scan + fields with an amber explanation and an **Open Settings** button (same style as Scan); `checking` disables Scan/Connect with a note; re-checks on every panel open and on `visibilitychange`, which also re-probes the desktop so a paired desktop reads unreachable once access is off. Strings added in en/es/fr/zh.

**Dialogs** — `components/ui/alert-dialog.tsx`, `components/ui/dialog.tsx`: `w-[calc(100%-2rem)]` + `rounded-lg` at every width, footer `gap-2` when stacked. Per-dialog overrides in `app-update-prompt.tsx` / `coding-mode-button.tsx` reduced to `max-w-md`.

**iOS build** — `bundleVersion` 5 (version 1.1.20) in `tauri.ios.conf.json`, `gen/apple/project.yml`, generated `Info.plist`. `PrivacyInfo.xcprivacy` force-added (gitignored, but referenced by the Xcode project — a fresh clone could not archive without it).

### Testing
- Rust: `cargo test --lib remote_code_client local_network` — 32 + 7 tests. New: late-answering desktop, address race, fast 401, dropped stream → continues, finished-while-away → completes from the desktop copy, foreground return → reconnect, coalescer catch-up, abandoned turns stopped / in-flight lifetime, prompt-aware probe (refusal-while-inactive retries, refusal-while-active answers, unanswered prompt ends at the window), host UDP smoke test. `cargo fmt`, clippy clean; `cargo check --target aarch64-apple-ios` clean.
- TS: `coding-mode-button` 94 tests (pairing single-call contract + hint; Local Network granted/denied/checking/re-check/desktop re-probe), `app-update-prompt` 12, ui/modals/chat suites green; typecheck + eslint clean.
- On device (iPhone 15 Pro, fresh installs): prompt at first open; Don't Allow → Open Settings → on → pairing UI returns; first scan pairs; approval-mode dialog; on/off in Settings reflected live.

### Notes for reviewers
- The Code-panel UI ships with the web platform: release apps load it from `os.ibl.ai`, so the panel changes appear there after the platform is deployed from this branch. The Rust-side fixes are in the app binary.
- LAN-pointed test builds must be **debug** builds: the dev-origin ACL capability in `lib.rs` is `#[cfg(debug_assertions)]`; a release build pointed at `http://192.168.x.x:3000` fails every invoke with "not allowed by ACL".
- The approval-mode dialog is per account (platform metadata `code_mode.permission_mode`) — a phone signed into an account that already answered on the Mac will not see it; that is existing behaviour, unchanged here.
- Still open (ops): `https://os.ibl.ai/.well-known/apple-app-site-association`; TestFlight upload of build 5.


null-crafter:
# PR 515 code review

**Head:** `4479dd5a` — "fix(tauri): adding fix for the ios code and UI responsive"
**Reviewed:** 2026-09-17 · follow-up to the [PR 513 rounds](https://github.com/iblai/os/pull/513#issuecomment-5697784226)

18 files, +2030/−134. The bulk is a new `local_network.rs` (551 lines) and
`remote_code_client.rs` (+1035). One thing to fix, one carried item that this PR makes
sharper, and some genuinely good work worth calling out.

---

## Carried findings — rechecked

| Finding | Result |
| --- | --- |
| **B2** — `os.ibl.ai` served no association file | **Fixed** ✅ |
| **N1** — `bundleVersion` pinned to `"2"` | **Addressed** ✅ |
| **S1** — `project.yml` build phase diverged | **Still open**, and sharper — below |
| **N2** — iOS `version: "1.1.20"` vs `tauri.conf.json`'s `0.95.19` | Unchanged |
| **S2** — `build.rs` cites `scripts/android-release.sh` | Still absent |

**B2 verified by content, not status code:**

```
$ curl -s https://os.ibl.ai/.well-known/apple-app-site-association
{ "applinks": { "details": [ { "appID": "L4FWRM8W5Z.ai.ibl.mentorai",
                               "paths": [ "/sso-login", "/sso-login/*", "/mobile/*", "/test/*" ] } ] } }
```

Universal links on the new default host will now actually fire.

**N1** was bumped `"2"` → `"5"` in all three places it lives — `tauri.ios.conf.json`,
`project.yml` and `ibl-ai-os_iOS/Info.plist`. That is the right discipline when both a spec
and its generated artifact are tracked, which is what makes S1 stand out.

---

## Should fix

### F1. `Undetermined` is documented for a case that returns `Denied`

`LocalNetwork::Undetermined` says:

> No verdict: no Wi-Fi address to test with, **or the prompt was never answered inside the
> window**. Pairing may still be attempted; it raises the prompt itself.

But `probe` returns `Denied` when the window expires with the prompt still up, and its own
test pins that:

```rust
fn an_unanswered_prompt_ends_at_the_window() {
    let v = probe(Duration::from_millis(700), true, || Attempt::NoRoute, || false);
    assert_eq!(v, LocalNetwork::Denied);
}
```

This reaches the user. `coding-mode-button.tsx` branches on `denied` to *replace* the
pairing UI with an explanation and an Open Settings button:

```tsx
{localNetwork === 'denied' ? (
  // No pairing can work without Local Network access, so
  // the scan and the fields are pointless: say why, and
  // put the one useful action in their place.
```

So a user who simply has not tapped **Allow** within `PANEL_WINDOW` (20s) is told to go turn
on a permission they never turned off, and the scanner disappears while the prompt is still
on screen. `ensure` also calls `remember_asked()` for that verdict, recording an answer that
never happened.

Narrow in practice — the launch-time ask gets `LAUNCH_WINDOW` (90s) and usually resolves
first — but the fix is small: return `Undetermined` at the deadline, which is what the doc
says and what the UI already means by it, and skip `remember_asked()` on that path.

### S1 (carried). `project.yml` still runs the command `build-rust.sh` replaced

```
project.pbxproj   shellScript = "\"$SRCROOT/build-rust.sh\"";
project.yml       script: cargo tauri ios xcode-script -v --platform … ${ARCHS:?}
```

This PR makes the gap sharper rather than closing it: it *does* sync `CFBundleVersion`
"2" → "5" inside `project.yml`, so the spec is demonstrably maintained as a source of
truth — the build phase just was not brought along. A regeneration still restores the
"Connection refused" failure that `build-rust.sh` exists to avoid.

---

## Nits

**N3.** `InFlight::drop` does a whole-file `read_config` / `write_config` — the same shape
as the clobber fixed in `c53681d1`. It is synchronous and brief, so the window is far
smaller, and the consequence is benign: a lost removal leaves a stale `in_flight` entry
that costs one extra status probe on the next launch and is then forgotten. Flagging it
only because of this file's history. (It is also sync file IO on a runtime thread.)

**N4.** Three straight apostrophes in the new strings, against the typographic `’`
convention: `en` "can't", `fr` "l'accès" and "L'accès". No braces nearby so ICU is not at
risk — same class as the `fr` strings on PR 505.

**S2 (carried).** `build.rs` still points at `scripts/android-release.sh`, which is not in
the repo. Control: 27 files under `scripts/`, three `android-*`, none of them `-release`.

---

## Good

Worth saying explicitly, because most of this PR is careful work:

- **`local_network.rs` reverse-engineers an undocumented iOS behaviour and writes down the
  reasoning.** The first packet toward a LAN address raises the prompt; refusals come back
  `EHOSTUNREACH`; "prompt is up" is distinguished from "Don't Allow" by whether the app is
  active, since a system alert resigns active state. That is a real derivation, and the
  module header explains it well enough that the next person will not have to redo it.
- **`probe` takes `attempt` and `active` as closures**, so the whole state machine is unit
  tested rather than only reachable on a device.
- **Two concerns I formed while reading were already covered by tests.**
  `the_probe_aims_at_a_neighbour_never_at_itself` handles the case where the phone itself
  holds the `.1` address (`192.168.0.1` → probes `.2`), and
  `catch_up_to_emits_only_the_unseen_suffix` covers the diverged-copy branch.
- **Nothing is cached**, because the Settings switch flips live in both directions. Right
  call, and stated as the reason.
- **Permissions declared in both `permissions/default.toml` and `capabilities/default.json`** —
  the exact pairing that was broken in PR 505, where the capability existed but was never in
  the allow-list.
- **`PrivacyInfo.xcprivacy`** uses the correct required-reason codes (`E174.1` disk space,
  `C617.1` file timestamp), each tied to the actual call site in a comment.
- **The `remote_code_client.rs` work** — stream reconnection, catch-up from the desktop's
  copy, foreground resume, aborting turns a killed app abandoned — lands with a test each,
  including the awkward ones (a turn that finished while the stream was down; turns
  abandoned by a killed app).
- **i18n**: four keys, all four locales.

---

## Verification

- **F1:** `cargo test an_unanswered_prompt_ends_at_the_window` shows the current contract;
  changing the deadline to `Undetermined` should flip that assertion and the
  `remember_asked` path with it.
- **S1:** `rg -n 'script:' src-tauri/gen/apple/project.yml` against
  `rg -n 'shellScript' src-tauri/gen/apple/ibl-ai-os.xcodeproj/project.pbxproj`.
- **B2:** `curl https://os.ibl.ai/.well-known/apple-app-site-association` — now passes.

---

## Verdict

**Approved**, with caveats. No blocking defect: nothing crashes, loses data, or opens a
hole, and the new surface arrives with tests.

Caveats, none a reason to withhold:

- **F1** should be fixed — it tells a user the wrong thing about their own device state,
  and moves up if the 90s launch ask turns out not to resolve first in practice.
- **S1** before anyone regenerates the Xcode project.
- **N2** worth confirming as deliberate; **S2** and the nits at leisure.

This is an assistant review and does not replace the external reviewer's.

null-crafter:
# Rough edges hitting a first iOS build

Notes from taking this branch from a cold machine to a running app on the iOS Simulator.
Every item below was **re-verified against `4479dd5a`** (this PR's head) today — several
things we hit on our first attempt have since been fixed and are listed at the bottom so
they don't get re-raised.

**Scope caveat up front:** none of this originates in this PR. `gh pr diff 515 --name-only`
is 18 files, and none of them is the code involved — `get_app_url()` and the three URL
`unwrap`s are untouched here (the `lib.rs` hunks are `@@ -10`, `@@ -2429`, `@@ -3171`,
`@@ -3185`). `build-rust.sh` and the tracked `project.yml` came from `037ee3a5`, already on
main via #513. This is build/onboarding friction on the branch as a whole, not a review of
the diff.

---

## Blocking merge

**None.**

One candidate was investigated and cleared. A `cargo tauri ios dev` run failed with
`ld: library 'iconv' not found`, linking the *host* `build_script_build`
(`aarch64-apple-darwin`, `-mmacosx-version-min=11.0.0`) rather than the iOS target. The
theory was that `build-rust.sh` scopes `CFLAGS_<target>`/`CXXFLAGS_<target>` per-target
(lines 45–47) but leaves Xcode's exported `SDKROOT` (`…/iPhoneSimulator26.2.sdk`) applying
to every `cc` invocation, including host ones.

That theory is **wrong**. A/B:

```bash
cd src-tauri
SDKROOT=…/iPhoneSimulator26.2.sdk cargo build --lib   # exit 0, no iconv
touch build.rs
cargo build --lib                                     # exit 0, no iconv
```

Both succeed. The failure doesn't reproduce, and it's most likely local to that machine —
a `target/` populated before a macOS major upgrade (Darwin 24.6 → 27.0 mid-session). Noting
it only so the next person who sees it reaches for a clean rebuild instead of a code fix.

---

## Non-blocking

### N1 — `src-tauri/.env.local` is silently ignored for iOS builds

`build-rust.sh:26` does `: "${TAURI_DEV_URL:=https://os.ibl.ai}"` then exports it, and
`build.rs:dev_url_from_env_files()` checks `std::env::var("TAURI_DEV_URL")` **before** the
dotenv files. So a bare `cargo tauri ios dev` bakes the production URL into a *debug* build
and never consults `.env.local`.

The Tauri CLI doesn't inject `TAURI_DEV_URL` itself — confirmed two ways: no `TAURI_*` in
the Xcode build-phase environment dump, and before `build-rust.sh` existed the baked value
*was* the `.env.local` one. `make tauri-ios-dev` is unaffected because it exports the
variable in the shell first.

Net effect: edit `.env.local`, rebuild, and the app still loads `os.ibl.ai` with no
indication why. Worth either dropping the default or logging the resolved source.

*(Inferred from the code path plus the env dump — we never got a clean iOS build today to
read the `[ibl.ai] TAURI_DEV_URL at compile time:` line back.)*

### N2 — A scheme-less `TAURI_DEV_URL` aborts the app at launch

`src-tauri/src/lib.rs:2538`, `:2567`, `:2625` are `parse().unwrap()`. A value like
`192.168.50.45:3000` (no `http://`) yields `RelativeUrlWithoutBase`. The panic happens
inside the Tauri `setup` closure, which on iOS runs inside tao's `did_finish_launching` — an
Objective-C callback that cannot unwind — so it converts to `panic_cannot_unwind` → abort.

What you actually see is the app vanishing instantly and a backtrace whose visible portion
is all `panic_nounwind` / UIKit frames, with the real one-line cause scrolled off above it.
This cost us a full debugging cycle from a missing `http://`.

Not production-affecting — `build.rs` already refuses to bake a non-https URL into a release
build. Cheap fixes: normalize in `get_app_url()`, and warn from `build.rs`, which has the
precedent in its existing https check.

### N3 — `scripts/fix-xcode-cargo-path.sh` is obsolete and now harmful

The build phase is `"$SRCROOT/build-rust.sh"` (`project.pbxproj:305`), but that script
`sed`s the *old* `cargo tauri ios xcode-script` string into the project file.
`docs/tauri-ios-setup.md:30` and `:189` still tell you to run it. #513 replaced the
mechanism; this should probably have gone with it.

### N4 — `docs/tauri-ios-setup.md` is stale enough to mislead

`cd apps/mentor` at lines 29, 125 and 149 — there is no `apps/` directory in this repo. It
also states the fix script "runs automatically with `make` commands"; no Makefile target
references it.

### N5 — Dev-server port mismatch

`make tauri-ios-dev` and `make dev-mobile` use **3001**; `src-tauri/.env.local` uses
**3000**. Following the docs points the app at a port nothing is serving.

### N6 — `Accelerate.framework` survives only in the generated `project.yml`

`project.yml:81` has `- sdk: Accelerate.framework`, but `bundle.iOS` in `tauri.conf.json`
has only `developmentTeam` + `minimumSystemVersion`. Delete `project.yml`, re-run
`ios init`, and it's gone — the build then fails on undefined `_vDSP_vadd` / `_vDSP_vmul` /
`_vDSP_sve`, which reads as a Rust problem but isn't: `llama-cpp-sys-2` emits
`cargo:rustc-link-lib=framework=Accelerate`, and because the crate builds as a `staticlib`
those link directives never reach Xcode.

This is exactly how we lost it on the first attempt. `"frameworks": ["Accelerate"]` under
`bundle.iOS` regenerates it. Use the bare name — Tauri renders that as
`sdk: Accelerate.framework`; `"Accelerate.framework"` renders as nothing (verified by
regenerating both ways).

### N7 — `cargo tauri ios init` silently ignores an existing `project.yml`

It's a template file, so cargo-mobile2 skips it when present. Editing `bundle.iOS` and
re-running init produces a **zero diff** until you delete `project.yml` first — config edits
look like no-ops. Upstream behaviour, worth a line in the docs.

### N8 — Signing is pinned to one Apple team

`CODE_SIGN_IDENTITY = "iPhone Developer"` and `DEVELOPMENT_TEAM = "L4FWRM8W5Z"`
(`project.pbxproj` 330/331, 425/426). Fine in-house, but a contributor without that team
can't build even for the simulator without overriding it — despite the simulator needing
no profile at all (`PROVISIONING_PROFILE_REQUIRED = NO`, `ENTITLEMENTS_ALLOWED = NO`).

### N9 — No iOS entry point, no iOS CI

`package.json` has no `ios` script; iOS lives only in the Makefile. No workflow references
`tauri ios` or `xcodebuild`, so nothing in CI compiles the Rust for any target.

---

## Rechecked — already fixed since our first attempt

| Was | Now |
| --- | --- |
| `Accelerate.framework` missing → link failed on `_vDSP_*` | Present in tracked `project.yml:81` + committed pbxproj ✅ |
| `cargo: command not found` pressing Play in Xcode | `build-rust.sh` exports `$HOME/.cargo/bin` + Homebrew, and no longer needs the Tauri CLI as a parent ✅ |
| `get_app_url()` fell back to a hardcoded `http://192.168.1.46:3001` | Falls back to `https://os.ibl.ai` ✅ |
| `com.apple.developer.associated-domains` blocked free/personal teams | Entitlements file is now `<dict/>` ✅ |

null-crafter (approved):

